PR #8, opened 2022-09-15
Polluted location model

### LINKED ISSUE #6: Create PollutedLocation model class
## Description

Create `PollutedLocation` model class 

## Steps to take

1. Create a new class library project `Apsitvarkom.Models` where we'll store all model classes
2. Create a PollutedLocation model class with properties we decided in #4 

## Acceptance criteria

Project created and PollutedLocation class is created with the needed properties

6ar8nas:
## What was done

- Location class created
- PollutedLocation model created
- Unit tests project created and tests written for location related models

### DIFF
--- a/Apsitvarkom.Models/Enumerations.cs
+++ b/Apsitvarkom.Models/Enumerations.cs
@@ -8,7 +8,6 @@ public class Enumerations
     /// <summary>Enumeration type for pollution levels of the location.</summary>
     public enum LocationSeverityLevel
     {
-        None = 0,
         Low = 1,
         Moderate = 2,
         High = 3
--- a/Apsitvarkom.Models/Enumerations.cs
+++ b/Apsitvarkom.Models/Enumerations.cs
@@ -1,0 +1,17 @@
+﻿namespace Apsitvarkom.Models
+{
+    /// <summary>
+    /// Class for storing all enumerated types for models. 
+    /// </summary>
+    public class Enumerations
+    {
+        /// <summary>Enumeration type for pollution levels of the location.</summary>
+        public enum LocationSeverityLevel
+        {
+            None = 0,
+            Low = 1,
+            Moderate = 2,
+            High = 3
+        };
+    }
+}
--- a/Apsitvarkom.Models/Enumerations.cs
+++ b/Apsitvarkom.Models/Enumerations.cs
@@ -1,17 +1,16 @@
-﻿namespace Apsitvarkom.Models
+﻿namespace Apsitvarkom.Models;
+
+/// <summary>
+/// Class for storing all enumerated types for models.
+/// </summary>
+public class Enumerations
 {
-    /// <summary>
-    /// Class for storing all enumerated types for models. 
-    /// </summary>
-    public class Enumerations
+    /// <summary>Enumeration type for pollution levels of the location.</summary>
+    public enum LocationSeverityLevel
     {
-        /// <summary>Enumeration type for pollution levels of the location.</summary>
-        public enum LocationSeverityLevel
-        {
-            None = 0,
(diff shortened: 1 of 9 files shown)


Apsitvarkom.Models/Enumerations.cs
```diff
@@ -0,0 +1,16 @@
...
+
+/// <summary>
+/// Class for storing all enumerated types for models.
+/// </summary>
+public class Enumerations
+{
+    /// <summary>Enumeration type for pollution levels of the location.</summary>
+    public enum LocationSeverityLevel
+    {
+        None = 0,
```
BenasB: Do we need a `None` severity? not sure how this would be useful

Apsitvarkom.Models/Location.cs
```diff
@@ -0,0 +1,42 @@
...
+
+    /// <summary>
+    /// Latitude of the location. Cannot be less than -90.0 degrees or exceed 90.0 degrees.
+    /// </summary>
+    public double Latitude
+    {
+        get => _latitude;
+        set
+        {
+            if (value is >= -90.0 and <= 90.0)
```
BenasB, line 19: cool pattern matching:) didn't think about this

Apsitvarkom.Models/PollutedLocation.cs
```diff
@@ -0,0 +1,54 @@
...
+/// <summary>
+/// Class for storing captured polluted location records.
+/// </summary>
+public class PollutedLocation
+{
+    private int _progress;
+    private int _radius;
+
+    /// <summary>Unique identifier of the given record.</summary>
+    public string Id { get; init; } = null!;
```
BenasB: is this not of type `Guid`?

Apsitvarkom.UnitTests/Models/LocationTests.cs
```diff
@@ -0,0 +1,23 @@
+﻿using Apsitvarkom.Models;
+
+namespace Apsitvarkom.UnitTests.Models;
+
+public class LocationTests
+{
+    [Test]
+    [TestCase(-91.1, 14.7, "Latitude")]
```
BenasB: `nameof(Location.Latitude)` for the third argument in the `TestCase`? You usually want to avoid hardcoding strings

BenasB (changes requested):
We might replace the model validation with something more sophisticated later (we'll need more detailed API action responses on invalid data), but for now that is okay. Otherwise looks good, just some questions and suggestions

Apsitvarkom.UnitTests/Models/LocationTests.cs
6ar8nas: Good point. Will be changed here and in other cases.

Apsitvarkom.Models/PollutedLocation.cs
6ar8nas: Initially I used Guid, but then I thought we could save it as strings instead of Guid to simplify its storing while still maintaining its offered uniqueness. Will change it back.

Apsitvarkom.Models/Enumerations.cs
6ar8nas: Not sure - perhaps could be used for closed records. Reverted it for now, but I guess this could be a matter for discussions.

Apsitvarkom.Models/PollutedLocation.cs
BenasB: We will save it as strings when it's actually saved somewhere but if we have a type for it in C#, I don't see why we shouldn't use it

BenasB (approved):
LGTM, feel free to merge

PR #10, opened 2022-09-15
Remove unneeded api proejct reference

BenasB:
## What was done

We don't need this reference in UnitTesting project now


6ar8nas (approved):
Good catch

PR #12, opened 2022-09-16
Created polluted location reading from file

### LINKED ISSUE #2: Create polluted location reading from a file
## Description

Create a class that reads a json file and returns a list of polluted locations.

## Steps to take

1. Create a new class library project `Apsitvarkom.DataAccess` (or similar logical name) where we'll store all data accessing related stuff
2. Create an interface `IPollutedLocationRepository` with one method `GetAllPollutedLocations()`
3. Create an implementation of this interface called `PollutedLocationsFileStore` which implements `GetAllPollutedLocations` by reading a json file. In constructor, take in a file path to a a json file that the class will use as a source. Refer to [MS docs on JSON (de)serialization](https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-how-to?pivots=dotnet-6-0)
4. Create a test project `Apsitvarkom.UnitTests` (whatever testing framework you prefer).
5. Write a few unit tests for `PollutedLocationsFileStore`

## Acceptance criteria

Implemented `PollutedLocationsFileStore` class with unit testing

6ar8nas:
## What was done

- Apsitvarkom.DataAccess project created, implemented PollutedLocationDTO records reading from file.
- Implemented mapping from PollutedLocationDTO to PollutedLocation and vice-versa.
- Unit tests written.

### DIFF
--- a/Apsitvarkom.Api/Apsitvarkom.Api.csproj
+++ b/Apsitvarkom.Api/Apsitvarkom.Api.csproj
@@ -12,4 +12,8 @@
     <PackageReference Include="Swashbuckle.AspNetCore" Version="6.4.0" />
   </ItemGroup>
 
+  <ItemGroup>
+    <ProjectReference Include="..\Apsitvarkom.Models\Apsitvarkom.Models.csproj" />
+  </ItemGroup>
+
 </Project>
--- a/Apsitvarkom.Api/Program.cs
+++ b/Apsitvarkom.Api/Program.cs
@@ -7,6 +7,8 @@ builder.Services.AddControllers();
 builder.Services.AddEndpointsApiExplorer();
 builder.Services.AddSwaggerGen();
 
+builder.Services.AddAutoMapper(AppDomain.CurrentDomain.GetAssemblies());
+
 var app = builder.Build();
 
 // Configure the HTTP request pipeline.
--- a/Apsitvarkom.Api/Program.cs
+++ b/Apsitvarkom.Api/Program.cs
@@ -1,3 +1,5 @@
+using Apsitvarkom.Models.Mapping;
+
 var builder = WebApplication.CreateBuilder(args);
 
 // Add services to the container.
@@ -7,7 +9,7 @@ builder.Services.AddControllers();
 builder.Services.AddEndpointsApiExplorer();
 builder.Services.AddSwaggerGen();
 
-builder.Services.AddAutoMapper(AppDomain.CurrentDomain.GetAssemblies());
+builder.Services.AddAutoMapper(typeof(PollutedLocationProfile));
 
 var app = builder.Build();
 
--- a/Apsitvarkom.DataAccess/Apsitvarkom.DataAccess.csproj
+++ b/Apsitvarkom.DataAccess/Apsitvarkom.DataAccess.csproj
@@ -1,15 +1,18 @@
-<Project Sdk="Microsoft.NET.Sdk.Web">
+<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFramework>net6.0</TargetFramework>
-    <Nullable>enable</Nullable>
     <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
   </PropertyGroup>
 
   <ItemGroup>
     <PackageReference Include="AutoMapper" Version="11.0.1" />
     <PackageReference Include="AutoMapper.Extensions.Microsoft.DependencyInjection" Version="11.0.0" />
-    <PackageReference Include="Swashbuckle.AspNetCore" Version="6.4.0" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Apsitvarkom.Models\Apsitvarkom.Models.csproj" />
   </ItemGroup>
 
 </Project>
--- a/Apsitvarkom.DataAccess/Apsitvarkom.DataAccess.csproj
+++ b/Apsitvarkom.DataAccess/Apsitvarkom.DataAccess.csproj
@@ -1,15 +1,10 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFramework>net6.0</TargetFramework>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
   </PropertyGroup>
-
-  <ItemGroup>
-    <PackageReference Include="AutoMapper" Version="11.0.1" />
-    <PackageReference Include="AutoMapper.Extensions.Microsoft.DependencyInjection" Version="11.0.0" />
-  </ItemGroup>
 
   <ItemGroup>
     <ProjectReference Include="..\Apsitvarkom.Models\Apsitvarkom.Models.csproj" />
--- a/Apsitvarkom.DataAccess/IPollutedLocationDTORepository.cs
+++ b/Apsitvarkom.DataAccess/IPollutedLocationDTORepository.cs
@@ -1,0 +1,13 @@
+﻿using Apsitvarkom.Models.DTO;
+
+namespace Apsitvarkom.DataAccess;
+
+/// <summary>
+/// Interface for <see cref="PollutedLocationDTO" /> class data handling.
+/// </summary>
+public interface IPollutedLocationDTORepository
+{
+    /// <summary>Gets all <see cref="PollutedLocationDTO" /> records from the data source.</summary>
+    /// <returns>Array of all <see cref="PollutedLocationDTO" /> instances.</returns>
+    public PollutedLocationDTO[] GetAllPollutedLocations();
+}
--- a/Apsitvarkom.DataAccess/IPollutedLocationDTORepository.cs
+++ b/Apsitvarkom.DataAccess/IPollutedLocationDTORepository.cs
@@ -8,6 +8,6 @@ namespace Apsitvarkom.DataAccess;
 public interface IPollutedLocationDTORepository
 {
     /// <summary>Gets all <see cref="PollutedLocationDTO" /> records from the data source.</summary>
-    /// <returns>Array of all <see cref="PollutedLocationDTO" /> instances.</returns>
-    public PollutedLocationDTO[] GetAllPollutedLocations();
+    /// <returns><see cref="Enumerable" /> of <see cref="PollutedLocationDTO" /> instances.</returns>
+    public IEnumerable<PollutedLocationDTO> GetAllPollutedLocations();
 }
--- a/Apsitvarkom.DataAccess/PollutedLocationsDTOFileStore.cs
+++ b/Apsitvarkom.DataAccess/PollutedLocationsDTOFileStore.cs
@@ -1,0 +1,51 @@
+﻿using System.Text.Json;
+using Apsitvarkom.Models.DTO;
+
+namespace Apsitvarkom.DataAccess;
+
+/// <summary>
+/// Class for <see cref="PollutedLocationDTO" /> data handling from file.
+/// </summary>
+public class PollutedLocationsDTOFileStore : IPollutedLocationDTORepository
+{
+    /// <summary>JSON string type file contents to be parsed.</summary>
+    private readonly string _jsonString;
+    /// <summary>Default options for objects deserialization.</summary>
+    private readonly JsonSerializerOptions _options;
+
+    /// <summary>Constructor for the reader.</summary>
+    /// <param name="jsonString">JSON string data to be used for parsing.</param>
+    public PollutedLocationsDTOFileStore(string jsonString)
+    {
+        _jsonString = jsonString;
+        _options = new JsonSerializerOptions
+        {
+            AllowTrailingCommas = true,
+            PropertyNameCaseInsensitive = true
+        };
+    }
+
+    /// <inheritdoc />
+    public PollutedLocationDTO[] GetAllPollutedLocations()
+    {
+        var result = JsonSerializer.Deserialize<List<PollutedLocationDTO>>(_jsonString, _options);
+        return result is not null ? result.ToArray() : Array.Empty<PollutedLocationDTO>();
+    }
+
+    /// <summary>Static factory constructor for reader from file.</summary>
+    /// <param name="sourcePath">Relative location of the .json type source data file.</param>
+    public static PollutedLocationsDTOFileStore FromFile(string sourcePath)
+    {
+        if (sourcePath is null)
+            throw new ArgumentNullException(nameof(sourcePath), "Could not access file specified as the source path.");
+        using var reader = new StreamReader(sourcePath);
+        return FromContent(reader.ReadToEnd());
+    }
+
+    /// <summary>Static factory constructor for reader from JSON string.</summary>
+    /// <param name="contents">Contents to be parsed from JSON string.</param>
+    public static PollutedLocationsDTOFileStore FromContent(string contents)
+    {
+        return new PollutedLocationsDTOFileStore(contents);
+    }
+}
--- a/Apsitvarkom.DataAccess/PollutedLocationsDTOFileStore.cs
+++ b/Apsitvarkom.DataAccess/PollutedLocationsDTOFileStore.cs
@@ -10,12 +10,13 @@ public class PollutedLocationsDTOFileStore : IPollutedLocationDTORepository
 {
     /// <summary>JSON string type file contents to be parsed.</summary>
     private readonly string _jsonString;
+
     /// <summary>Default options for objects deserialization.</summary>
     private readonly JsonSerializerOptions _options;
 
     /// <summary>Constructor for the reader.</summary>
     /// <param name="jsonString">JSON string data to be used for parsing.</param>
-    public PollutedLocationsDTOFileStore(string jsonString)
+    private PollutedLocationsDTOFileStore(string jsonString)
     {
         _jsonString = jsonString;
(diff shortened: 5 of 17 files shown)


BenasB:
I haven't looked at the changed files yet, but I can see that this PR has 1 commit. Since this was a pretty big task, try to do incremental commits in the future to get a more gradual flow of updates. Nonetheless, thanks for doing this, I will review it shortly

Apsitvarkom.Api/Program.cs
```diff
@@ -7,6 +7,8 @@
 builder.Services.AddEndpointsApiExplorer();
 builder.Services.AddSwaggerGen();
 
+builder.Services.AddAutoMapper(AppDomain.CurrentDomain.GetAssemblies());
```
BenasB: As per [MS docs](https://learn.microsoft.com/en-us/dotnet/api/system.appdomain.getassemblies?view=net-6.0), `AppDomain.CurrentDomain.GetAssemblies()` will return all loaded assemblies. Let's be more specific and get only the assembly where our mappings are `Apsitvarkom.Models/Mapping` (which is most probably the `Apsitvarkom.Model` assembly). As I can see from the [automapper docs](https://docs.automapper.org/en/stable/Dependency-injection.html#asp-net-core), there is an overload for `AddAutoMapper` that can take a type and it will scan the assembly of that type.

So a better choice that will achieve the same here would be `builder.Services.AddAutoMapper(typeof(PollutedLocationProfile));`

Apsitvarkom.DataAccess/IPollutedLocationDTORepository.cs
```diff
@@ -0,0 +1,13 @@
...
+namespace Apsitvarkom.DataAccess;
+
+/// <summary>
+/// Interface for <see cref="PollutedLocationDTO" /> class data handling.
+/// </summary>
+public interface IPollutedLocationDTORepository
+{
+    /// <summary>Gets all <see cref="PollutedLocationDTO" /> records from the data source.</summary>
+    /// <returns>Array of all <see cref="PollutedLocationDTO" /> instances.</returns>
+    public PollutedLocationDTO[] GetAllPollutedLocations();
```
BenasB: From my experience, we return a more generic type instead of an array when dealing with collections. Returning `IEnumerable<PollutedLocationDTO>` instead of `PollutedLocationDTO[]` would make more sense. `IEnumerable` is a very generic interface that basically says "this is a list, you can iterate over it", see more about prefering `IEnumerable`s instead of arrays [in this SO question](https://stackoverflow.com/questions/5156561/should-i-favour-ienumerablet-or-arrays).

Apsitvarkom.DataAccess/PollutedLocationsDTOFileStore.cs
```diff
@@ -0,0 +1,51 @@
...
+            AllowTrailingCommas = true,
+            PropertyNameCaseInsensitive = true
+        };
+    }
+
+    /// <inheritdoc />
+    public PollutedLocationDTO[] GetAllPollutedLocations()
+    {
+        var result = JsonSerializer.Deserialize<List<PollutedLocationDTO>>(_jsonString, _options);
+        return result is not null ? result.ToArray() : Array.Empty<PollutedLocationDTO>();
```
BenasB: Would `return result?.ToArray() ?? Array.Empty<PollutedLocationDTO>()` work?

Apsitvarkom.DataAccess/PollutedLocationsDTOFileStore.cs
```diff
@@ -0,0 +1,51 @@
...
+public class PollutedLocationsDTOFileStore : IPollutedLocationDTORepository
+{
+    /// <summary>JSON string type file contents to be parsed.</summary>
+    private readonly string _jsonString;
+    /// <summary>Default options for objects deserialization.</summary>
+    private readonly JsonSerializerOptions _options;
+
+    /// <summary>Constructor for the reader.</summary>
+    /// <param name="jsonString">JSON string data to be used for parsing.</param>
+    public PollutedLocationsDTOFileStore(string jsonString)
```
BenasB: You have two static factory methods, so this constructor shouldn't be used outside of the class, declare it `private`

Apsitvarkom.DataAccess/PollutedLocationsDTOFileStore.cs
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    /// <summary>Static factory constructor for reader from file.</summary>
+    /// <param name="sourcePath">Relative location of the .json type source data file.</param>
+    public static PollutedLocationsDTOFileStore FromFile(string sourcePath)
+    {
+        if (sourcePath is null)
+            throw new ArgumentNullException(nameof(sourcePath), "Could not access file specified as the source path.");
+        using var reader = new StreamReader(sourcePath);
+        return FromContent(reader.ReadToEnd());
```
BenasB: Consider using `File.ReadAllText()` to do this simple file read. [Reference](https://stackoverflow.com/a/3545407)

Apsitvarkom.Models/Apsitvarkom.Models.csproj
```diff
@@ -1,9 +1,19 @@
...
   </PropertyGroup>
 
+  <ItemGroup>
+    <Folder Include="DTO\" />
+    <Folder Include="Mapping\" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="AutoMapper" Version="11.0.1" />
+    <PackageReference Include="AutoMapper.Extensions.Microsoft.DependencyInjection" Version="11.0.0" />
```
BenasB: `AutoMapper.Extensions.Microsoft.DependencyInjection` is probably not needed here

Apsitvarkom.Models/DTO/LocationDTO.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace Apsitvarkom.Models.DTO;
+
+/// <summary>
+/// Data Transfer Object for <see cref="Location" />
+/// </summary>
+public class LocationDTO
```
BenasB: Consider making this a struct? I think one of the requirements for lab 1 is that we know how to create structs 😄

BenasB: It would make sense here, since we only have 2 primitive fields

Apsitvarkom.DataAccess/PollutedLocationsDTOFileStore.cs
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    /// <summary>Static factory constructor for reader from JSON string.</summary>
+    /// <param name="contents">Contents to be parsed from JSON string.</param>
+    public static PollutedLocationsDTOFileStore FromContent(string contents)
+    {
+        return new PollutedLocationsDTOFileStore(contents);
+    }
```
BenasB: Lambda expression for this method?

Apsitvarkom.DataAccess/PollutedLocationsDTOFileStore.cs
```diff
@@ -0,0 +1,51 @@
...
+        var result = JsonSerializer.Deserialize<List<PollutedLocationDTO>>(_jsonString, _options);
+        return result is not null ? result.ToArray() : Array.Empty<PollutedLocationDTO>();
+    }
+
+    /// <summary>Static factory constructor for reader from file.</summary>
+    /// <param name="sourcePath">Relative location of the .json type source data file.</param>
+    public static PollutedLocationsDTOFileStore FromFile(string sourcePath)
+    {
+        if (sourcePath is null)
+            throw new ArgumentNullException(nameof(sourcePath), "Could not access file specified as the source path.");
```
BenasB: [MS docs](https://learn.microsoft.com/en-us/dotnet/api/system.io.streamreader.-ctor?view=net-6.0#system-io-streamreader-ctor(system-string)) say that `new StreamReader(string)` Already throws an exception indicating the same thing (and more). I am not sure if this null check is useful, since we'd get an exception indicating this either way.

Apsitvarkom.DataAccess/Apsitvarkom.DataAccess.csproj
```diff
@@ -0,0 +1,18 @@
...
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="AutoMapper" Version="11.0.1" />
+    <PackageReference Include="AutoMapper.Extensions.Microsoft.DependencyInjection" Version="11.0.0" />
+  </ItemGroup>
```
BenasB: Both of these are probably not needed in `Apsitvarkom.DataAccess` project. Rule of thumb is to only include packages where you use it (I don't think we're using it in this project).

Apsitvarkom.Api/Apsitvarkom.Api.csproj
```diff
@@ -7,7 +7,9 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Swashbuckle.AspNetCore" Version="6.2.3" />
+    <PackageReference Include="AutoMapper" Version="11.0.1" />
+    <PackageReference Include="AutoMapper.Extensions.Microsoft.DependencyInjection" Version="11.0.0" />
```
BenasB, line 11: Is it possible to have `AutoMapper` package in `Apsitvarkom.Models` project since there's where all of our mappings will be located? And of course keep `AutoMapper.Extensions.Microsoft.DependencyInjection` in `Apsitvarkom.Api`, since it's needed to register mapping profiles to DI.

Apsitvarkom.UnitTests/DataAccess/PollutedLocationsFileStoreTests.cs
```diff
@@ -0,0 +1,103 @@
+﻿using Apsitvarkom.DataAccess;
+using static Apsitvarkom.Models.Enumerations;
+
+namespace Apsitvarkom.UnitTests.DataAccess;
+
+public class PollutedLocationsFileStoreTests
+{
+    [Test]
+    [TestCase("9719d4ef-5cde-4370-a510-53af84bdede2", -181.12311, LocationSeverityLevel.High, "2015-05-16T05:50:06", 100)]
+    public void GetAllPollutedLocationsDTO_SomePropertiesMissing_DeserializingSuccessful_MissingPropertiesSetToNull(string id, double longitude, LocationSeverityLevel severity, string creationTime, int progress)
```
BenasB, line 10: This got me thinking. If I understand correctly, if a json object is missing some properties, the deserialization will still succeed, but some of the fields will be null. In the DTO, we say that these properties `= null!` which is not true in this case. I think it would be best to have DTO properties nullable. E.g. transform `public string Id { get; init; } = null!;` into `public string? Id { get; init; }`

BenasB: Unless there's a way to make the `JsonSerializer.Deserialize` throw an exception if it does not find **all** of the fields. But I am not sure if that's possible.

BenasB: When the DTO properties are nullable, check how automapper handles when mapping `PollutedLocationDTO` -> `PollutedLocation` and some of the fields are null. It would be best if the `Map` method throws in that case. It would make sense to add a unit test for that case in `Apsitvarkom.UnitTests/Models/Mapping/PollutedLocationMappingTests.cs`

BenasB (changes requested):
Overall looking good, although please address the comments and push changes if needed. We may need to make the `IPollutedLocationRepository` methods `async` (database queries will definitely be async), but let's do this one step at a time.

Apsitvarkom.Api/Apsitvarkom.Api.csproj
6ar8nas: Yes, that is possible. I have excessively left the package in the DataAccess project, as I wasn't sure whether we should return already mapped results or not, so I will revert it.

Apsitvarkom.DataAccess/Apsitvarkom.DataAccess.csproj
6ar8nas: Discussed this below. Will be sorted out.

> Yes, that is possible. I have excessively left the package in the DataAccess project, as I wasn't sure whether we should return already mapped results or not, so I will revert it.

Apsitvarkom.DataAccess/PollutedLocationsDTOFileStore.cs
6ar8nas: Nice find. Didn't check into the docs enough. Will change it as File.ReadAllText() also throws if the path is null.

Apsitvarkom.DataAccess/PollutedLocationsDTOFileStore.cs
6ar8nas: Thanks, accidentally left this out after I started using the static factory methods.

Apsitvarkom.DataAccess/PollutedLocationsDTOFileStore.cs
6ar8nas: It would. I had doubts whether this would be as easy to read though, but I believe now that we change it to enumerable it will be even better.

Apsitvarkom.DataAccess/IPollutedLocationDTORepository.cs
6ar8nas: Agreed, similar experiences on this one.

Apsitvarkom.Api/Program.cs
6ar8nas: [This](https://stackoverflow.com/a/72161847) comment suggests the same. Done.

Apsitvarkom.UnitTests/DataAccess/PollutedLocationsFileStoreTests.cs
6ar8nas: > Unless there's a way to make the `JsonSerializer.Deserialize` throw an exception if it does not find **all** of the fields. But I am not sure if that's possible.

https://github.com/dotnet/runtime/issues/37483 - seems like there is no such way as of yet.

> It would be best if the Map method throws in that case. It would make sense to add a unit test for that case in Apsitvarkom.UnitTests/Models/Mapping/PollutedLocationMappingTests.cs

Currently it doesn't work like that. I believe we should create a new issue for that.

Apsitvarkom.Models/DTO/LocationDTO.cs
```diff
@@ -3,11 +3,13 @@
 /// <summary>
 /// Data Transfer Object for <see cref="Location" />
 /// </summary>
-public class LocationDTO
+public struct LocationDTO
 {
+    public LocationDTO(double? longitude, double? latitude) => (Longitude, Latitude) = (longitude, latitude);
```
BenasB: Why is this constructor needed? I think we've decided to create these types of classes like this:
```csharp
new LocationDTO(){
  longitude: 100,
  latitude: 100
}
```

Apsitvarkom.Models/Mapping/PollutedLocationProfile.cs
```diff
@@ -15,9 +15,9 @@ public PollutedLocationProfile()
         // PollutedLocationDTO to PollutedLocation
         CreateMap<PollutedLocationDTO, PollutedLocation>()
             .ForMember(dest => dest.Id, opt => opt
-                .MapFrom(src => Guid.Parse(src.Id)))
+                .MapFrom(src => Guid.Parse(src.Id ?? string.Empty)))
```
BenasB, line 18: As discussed, `src.Id` will be validated to not be null after we put explicit validation

Apsitvarkom.Models/Mapping/PollutedLocationProfile.cs
```diff
@@ -15,9 +15,9 @@ public PollutedLocationProfile()
...
             .ForMember(dest => dest.Spotted, opt => opt
-                .MapFrom(src => DateTime.Parse(src.Spotted, CultureInfo.InvariantCulture)));
+                .MapFrom(src => DateTime.Parse(src.Spotted ?? string.Empty, CultureInfo.InvariantCulture)));
```
BenasB, line 20: Same for `src.Spotted`

BenasB (changes requested):
Small things

Apsitvarkom.UnitTests/DataAccess/PollutedLocationsFileStoreTests.cs
BenasB: Let's create a new task for that: adding explicit validation for DTO objects. I'll do that and add it to the project board

Apsitvarkom.Models/Mapping/PollutedLocationProfile.cs
6ar8nas: Possible null reference warning. Even though we will put it through validation, I'll keep it here for now.

Apsitvarkom.Models/Mapping/PollutedLocationProfile.cs
6ar8nas: Possible null reference warning. Even though we will put it through validation, I'll keep it here for now.

BenasB (approved):
LGTM now, feel free to merge

6ar8nas:
Possible improvements:
Making IPollutedLocationRepository methods async #13  
Validating deserialized DTO objects for mapping #14

PR #15, opened 2022-09-17
Make IPollutedLocationDTORepository async

### LINKED ISSUE #13: Make IPollutedLocationDTORepository async
## Description

Make `IPollutedLocationDTORepository` methods async since in a real scenario we'll be querying a database which will definitely use an async interface. If you're not familiar with asynchronous programming, see [MS Docs](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/async/) as a starting point.

## Steps to take

1. In [ `IPollutedLocationDTORepository`](https://github.com/vu-vibedosa/apsitvarkom-be/blob/dev/Apsitvarkom.DataAccess/IPollutedLocationDTORepository.cs) wrap the method return types in `Task<T>`  and add `Async` to the end of method names (this is the usual convention).
2. Changing the interface will require changes in all its implementations. Now we only have [`PollutedLocationsDTOFileStore`](https://github.com/vu-vibedosa/apsitvarkom-be/blob/dev/Apsitvarkom.DataAccess/PollutedLocationsDTOFileStore.cs) (btw it would make sense to rename this to `Repository` instead of `Store`, idk why I came up with that initially)
    1. This will require to use [`JsonSerializer.DeserializeAsync`](https://docs.microsoft.com/en-us/dotnet/api/system.text.json.jsonserializer.deserializeasync?view=net-6.0)

## Acceptance criteria
All interface `IPollutedLocationDTORepository` methods are async and its implementations are adjusted accordingly.

6ar8nas:
## What was done

- Renamed `PollutedLocationsDTOFileStore` to `PollutedLocationsDTOFileRepository`
- Changed `IPollutedLocationsDTORepository` methods to be async and adjusted its `PollutedLocationsDTOFileRepository` implementation accordingly.


BenasB (changes requested):
Implementation looking good.

Although usage in tests is not in a way that you usually use `async` methods. Checking [NUnit docs](https://docs.nunit.org/articles/nunit/writing-tests/attributes/test.html) says that you can do an async test:

```csharp
// A simple async test
[Test]
public async Task AddAsync()
{ /* ... */ }
```

And then you actually `await` the call instead of `.Result`ing it and running it synchronously (running it synchronously doesn't matter a lot in this scenario, but you shouldn't do it if you can here).

Also you don't dispose the repository instance in the tests, use a [`using`](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-statement) when you're creating the `dataManager`

And one more thing, I think it would be good to have an actual test that reads json from a file and not from a string. You can create a json file in the project and check "Copy always" to it – this way the file will end up next to your compiled DLL and you will be able to reference it (just by specifying a name, relatively) easily from the test.

6ar8nas:
Thanks for the remarks.

> And one more thing, I think it would be good to have an actual test that reads json from a file and not from a string. You can create a json file in the project and check "Copy always" to it – this way the file will end up next to your compiled DLL and you will be able to reference it (just by specifying a name, relatively) easily from the test.

Regarding this - wouldn't it be better to create the file in OneTimeSetUp and delete in OneTimeTearDown, so that when the tests would be running, we would have that file existing and could reference it as easily from the test, but we would clean up every time after the test session ends?

BenasB:
I do not see why we would write additional code to handle string writing to a file. We can include a real json file as build content that will be copied over to the build directory automatically when compiling the unit test project. Here's an example:

image

Then in the test you would just need to specify the file as a relative path to the test dll (in the example that would be `./DataAccess/PollutedLocationsMock.json`)

Apsitvarkom.UnitTests/DataAccess/PollutedLocationsFileRepositoryTests.cs
```diff
@@ -1,13 +1,27 @@
...
+
+    // Existing mock data file, containing three valid instances with unique property values.
+    private string m_validDataSourcePath;
+
+    [OneTimeSetUp]
+    public void OneTimeSetUp()
+    {
+        m_validDataSourcePath = Path.Combine("DataAccess", "PollutedLocationDTOMockValid.json");
+        m_invalidDataSourcePath = Path.Combine("DataAccess", "PollutedLocationDTOMockInvalid.json");
+    }
```
BenasB: I believe you don't even have to have this `OneTimeSetUp` here. 

`private static readonly string InvalidDataSourcePath = Path.Combine("DataAccess", "PollutedLocationDTOMockValid.json")`

I think should work

BenasB (changes requested):
One more comment, but good job overall

BenasB (approved):
LGTM, feel free to merge

PR #17, opened 2022-09-18
Hotfix unit testing jsonString construction culture settings

BenasB:
## What was done

Fix `ToString` for doubles to ignore culture settings when constructing jsonString


6ar8nas (approved):
Didn't see this being an issue when I wrote this. That's nice, that we are using different localization settings.

Issue #4, opened 2022-09-14
Discuss and determine polluted location model


BenasB:
Data type
PollutedLocation
```
Id: guid string
Location: coordinates (latitude and longitude)
Radius: integer >= 1 (in meters)
Severity: "Low" | "Moderate" | "High"
Spotted: Timestamp
Progress: 0-100 in %
Notes: string
```

JSON example:
```json
{
    "id": "5be2354e-2500-4289-bbe2-66210592e17f",
    "location": {
        "longitude": -78.948237,
        "latitude": 35.929673
    },
    "radius": 10,
    "severity": "Moderate",
    "spotted": "2022-09-14T17:35:23Z",
    "progress": 0,
    "notes": "Lorem ipsum"
}
```

Fields to be added later:
- Photo
- SpottedBy

6ar8nas:
Edited longitude and latitude - value should not be in quotation marks.

PR #18, opened 2022-09-19
PollutedLocation Controller handling Actions

### LINKED ISSUE #7: Create PollutedLocation controller
## Prerequisites

This task requires #6 to be finished, do not start it before that.

## Description

Create a simple `PollutedLocation` controller that will take care of all requests relating to polluted locations.

## Steps to take

1. Create the controller class in `Apsitvarkom.Api/Controllers/`.
2. Create one GET action/endpoint at route `/PollutedLocation/All` that returns a constant list (`IEnumerable`) of `PollutedLocationDTO` objects (initialize them with some fake data)

## Acceptance criteria

Endpoint is reachable at `/PollutedLocation/All` and this action is seen in swagger. This action returns a json list containing objects we described in #4

XxDoMaSxX:
## What was done

-Created PollutionLocationController 
-Created Actions for Controller

Created Actions : Create, Update, Delete, Get All, Get By Id


Apsitvarkom.Api/Controllers/PollutedLocationController.cs
```diff
@@ -0,0 +1,115 @@
+﻿using Apsitvarkom.Models;
+using Apsitvarkom.Models.DTO;
+using Microsoft.AspNetCore.Mvc;
+
+
```
6ar8nas, line 6: unnecessary new line

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
```diff
@@ -0,0 +1,115 @@
...
+namespace Apsitvarkom.Api.Controllers  
+{
```
6ar8nas, line 10: You can write ```namespace Apsitvarkom.Api.Controllers;``` and omit the curly brackets.

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
```diff
@@ -0,0 +1,115 @@
...
+    [ApiController]
+    [Route("/api/[controller]")]
+
```
6ar8nas: new line

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
```diff
@@ -0,0 +1,115 @@
...
+                },
+                Radius = 10,
+                Severity = "MODERATE",
+                Spotted  =  DateTime.Now.ToString("yyyy-MM-dd HH:mm:ss.ttt"),
+                Progress = 15,
+                Notes = "This is test notes"
+
+            }
+        };
+        /*@Parameters : Object Id
```
6ar8nas: Check out [this](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/documentation-comments) documentation comments specification for C# or take a look at [our own code example](https://github.com/vu-vibedosa/apsitvarkom-be/blob/dev/Apsitvarkom.DataAccess/PollutedLocationsDTOFileRepository.cs).

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
```diff
@@ -0,0 +1,115 @@
...
+
+            }
+        };
+        /*@Parameters : Object Id 
+         * @Returns : List of objects that are still in List
+         * @Action : Deleted object from list
+         */
+
+        [HttpGet("{id}")]
+        public async Task<ActionResult<PollutedLocationDTO>> GetById(string id)
```
6ar8nas: You specify the method as async Task, but do not use await anywhere in the method. This conversion may be a new gradual increment to this task, consider not using ```async Task<>``` for now.

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
```diff
@@ -0,0 +1,115 @@
...
+            {
+                return BadRequest("Polluted location is not found - wrong id provided or id is not provided at all");
+            }
+            return Ok(PollutedLocation);
+        }
+        /*@Parameters : 
+         * @Returns : List of all objects
+         * @Action : Returns all list of objects 
+         */
+        [HttpGet]
```
6ar8nas: The task specified the endpoint to be reachable at `/PollutedLocation/All`.

6ar8nas (changes requested):
Try to refer more to the updated [ExampleController](https://github.com/vu-vibedosa/apsitvarkom-be/blob/dev/Apsitvarkom.Api/Controllers/ExampleController.cs). It includes a constructor which you're missing and more.

There are a few whitespace misuses (new lines where they are not needed, new lines missing where I would opt to use them). You can also check out [this](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/documentation-comments) documentation comments specification for C# or take a look at [our own code example](https://github.com/vu-vibedosa/apsitvarkom-be/blob/dev/Apsitvarkom.DataAccess/PollutedLocationsDTOFileRepository.cs) for future use.

Apsitvarkom.Api/package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
```
6ar8nas: I am uncertain and curious whether we really need this file. Maybe @BenasB could contribute on this matter?

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
```diff
@@ -0,0 +1,115 @@
...
+
+
+namespace Apsitvarkom.Api.Controllers  
+{
+    [ApiController]
+    [Route("/api/[controller]")]
+
+    public class PollutedLocationController : ControllerBase
+    {
+        /*temporary list PollutedLocations. Latter on we will have stored data in file or database. Then we will create
```
6ar8nas: I would suggest this to be deleted, as when our team is not big, we can track all our changes in our [project](https://github.com/orgs/vu-vibedosa/projects/1). If you were considering this comment to be useful for the lecturer, we can update him verbally.

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
```diff
@@ -0,0 +1,115 @@
...
+        static DataContext object _context and in constructor
+        declare _context value to DataContext object
+        */
+        
+        private static IEnumerable<PollutedLocationDTO> PollutedLocations = new List<PollutedLocationDTO>
+        {
+
```
6ar8nas: why have two new lines instead of one? (16-17)

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
```diff
@@ -0,0 +1,115 @@
...
+        }
+        /*@Parameters : Object Id 
+         * @Returns : List of objects that are still in List
+         * @Action : Deleted object from list
+         */
+
+        [HttpDelete("{id}")]
+        public async Task<ActionResult<PollutedLocationDTO>> Delete(string id)
+        {
+            var FoundPollutionLocation = PollutedLocations.FirstOrDefault(PollutedPlace => PollutedPlace.Id == id);
```
6ar8nas: consider using `camelCase` for local variables.

Some concise information on variable naming conventions [here](https://github.com/ktaranov/naming-convention/blob/master/C%23%20Coding%20Standards%20and%20Naming%20Conventions.md).

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
```diff
@@ -0,0 +1,115 @@
...
+            return Ok(PolutedLocation);
+        }
+
+        /*@Parameters : Object 
+         * @Returns : Updated Object
+         * @Action : Updates given object by its Id 
+         */
+
+        [HttpPut]
+        public async Task<ActionResult<List<PollutedLocation>>> Update(PollutedLocationDTO PollutedLocation)
```
6ar8nas, line 84: why do we request the user to provide us with a PollutedLocationDTO instance, when in all other cases we use Id?

6ar8nas:
Food for thought, not related to this task's scope - looking at lines 
```return BadRequest("Polluted location is not found - wrong id provided or id is not provided at all");```
maybe we should consider creating generic error handling messages that wouldn't differ majorly from one to another in different controllers?

File: Apsitvarkom.Api/package-lock.json (hunk 7 above)
Comment (XxDoMaSxX) [reply Apsitvarkom.Api/package-lock.json]: Accidently added. Removed on next commit

**Comment by XxDoMaSxX** [comment]
> Food for thought, not related to this task's scope - looking at lines `return BadRequest("Polluted location is not found - wrong id provided or id is not provided at all");` maybe we should consider creating generic error handling messages that wouldn't differ majorly from one to another in different controllers?

Good Point, latter on I guess we will need it anyways to flash errors. For a now I leave those errors, will change when error logic will be implemented. What do you think about that ?

File: Apsitvarkom.Api/package-lock.json (hunk 7 above)
Comment (6ar8nas) [reply Apsitvarkom.Api/package-lock.json]: You didn't delete it fully, just added it as a 'remove' tag to Apsitvarkom.Api.csproj. Delete from both places now.

EDIT: resolving it as a newer comment regarding this has been created.

File: Apsitvarkom.Api/Controllers/PollutedLocationController.cs [hunk 12]
```diff
@@ -0,0 +1,95 @@
...
+        return Ok(PolutedLocation);
+    }
+
+    [HttpPut]
+    public ActionResult<List<PollutedLocation>> Update(PollutedLocationDTO PollutedLocation)
+    {
+
+        var list = pollutedLocations.ToList();
+        if (list.Remove(list.Where(predicate: x => x.Id == PollutedLocation.Id).FirstOrDefault()))
+        {
```
6ar8nas: Solve these warnings.

EDIT: As discussed below by Benas, it is no longer in this task's scope.

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
```diff
@@ -0,0 +1,95 @@
...
+        var pollutedLocation = pollutedLocations.FirstOrDefault(PollutedPlace => PollutedPlace.Id == id);
+        if (pollutedLocation == null)
+        {
+            return BadRequest("Polluted location is not found - wrong id provided or id is not provided at all");
+        }
+        return Ok(pollutedLocation);
+    }
+
+    [HttpGet]
+    [Route("/[controller]/All")]
```
6ar8nas: on line 37 you use HttpGet("{id}"), but here you use HttpGet and then Route. Why not have it more consistently?

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
```diff
@@ -0,0 +1,95 @@
...
+    public ActionResult<List<PollutedLocationDTO>> GetAll()
+    {
+
+        return Ok(pollutedLocations);
+    }
+
+    [HttpPost]
+    public ActionResult<IEnumerable<PollutedLocationDTO>> Insert(PollutedLocationDTO PolutedLocation)
+    {
+
```
6ar8nas: Why do we need new lines before or after code functionality?

```
public void Method()
{
/* new lines here*/
<First sentence in the method>
...
<Last sentence of the method>
/* or new lines here*/
}
```

let's end it with the last sentence without additional whitespace:

```
public void Method()
{
<First sentence in the method>
...
<Last sentence of the method>
}

**Review by 6ar8nas — CHANGES_REQUESTED** [review]

File: Apsitvarkom.Api/Controllers/PollutedLocationController.cs (hunk 10 above)
Comment (6ar8nas) [reply Apsitvarkom.Api/Controllers/PollutedLocationController.cs]: You have also changed one static variable. If you'd check the documentation linked above, it suggests having them named in PascalCase

File: Apsitvarkom.Api/Controllers/PollutedLocationController.cs (hunk 13 above)
Comment (6ar8nas) [reply Apsitvarkom.Api/Controllers/PollutedLocationController.cs]: As you have asked for clarification on this one, in some places you use `[HttpGet("{id}")]` and in this case you use `[HttpGet]` together with `[Route("/[controller]/All")]`. Why use two different routing approaches, instead of using `[HttpGet("All")]`.

EDIT: resolving it as a newer comment regarding this has been created.

**Review by BenasB — CHANGES_REQUESTED** [review]
Hi everyone,

the main drawback with this PR is that you're **coupling (mixing) data access layer with the API layer**. To prevent this, we are making use of the [repository pattern](https://www.c-sharpcorner.com/article/repository-design-pattern-in-asp-net-mvc/) (important theory until "Implementation" section) which we started creating in #2. We should continue to do so and in fact are evolving it with future tasks, such as #19.

We're doing it incrementally (step by step) to make development and reviewing easier – preferring to work in small steps which take less time rather than in large chunks that take a longer time.

At the time of creating this task (#7) we didn't have the repository interface, so we couldn't implement [CRUD](https://en.wikipedia.org/wiki/Create,_read,_update_and_delete) actions as of then (we can't do that yet right now too, because the repository interface is still missing some methods, we're working on it).

So, my suggestion is:
- Right now, stick to the original #7 description and leave **only** `GetAll` action which would look something like (of course remove anything else that is unused):
```
public ActionResult<List<PollutedLocationDTO>> GetAll()
{
    var fakeData = new List<PollutedLocationDTO>() {
        // Create some fake data here
    };

    return Ok(fakeData);
}
``` 
- Keep your code changes in a separate branch locally, so when we are ready to implement other CRUD actions, you would be able to take those tasks and maybe reuse some of the code.

This is only what the task described to do. And we will add other actions task by task.

How does this compromise sound to all of you? 
@XxDoMaSxX  @6ar8nas  @BaVilius

**Comment by 6ar8nas** [comment]
> How does this compromise sound to all of you? @XxDoMaSxX @6ar8nas @BaVilius

I also didn't know exactly how to address the task scope - that seems valid.

> Keep your code changes in a separate branch locally, so when we are ready to implement other CRUD actions, you would be able to take those tasks and maybe reuse some of the code.

Just huge emphasis on this ^.

File: Apsitvarkom.Api/Controllers/PollutedLocationController.cs [hunk 15]
```diff
@@ -0,0 +1,95 @@
+﻿using Apsitvarkom.Models;
+using Apsitvarkom.Models.DTO;
+using Microsoft.AspNetCore.Mvc;
+namespace Apsitvarkom.Api.Controllers;
```
6ar8nas, line 5: According to what we have been doing in other files, a new line would be preferred here before the namespace. That's up for a discussion if You think this way is better though, but we would ideally have it consistently.

BaVilius:
> How does this compromise sound to all of you? @XxDoMaSxX @6ar8nas @BaVilius

Sounds good to me, leaving some of CRUD actions unimplemented for now will make this PR complete faster while also increasing the chance that further implementations will require less modifications. If @XxDoMaSxX leaves these changes local for now,  there really is nothing to lose afaik.

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
```diff
@@ -0,0 +1,55 @@
...
+{
+    private readonly ILogger<PollutedLocationController> _logger;
+
+    public PollutedLocationController(ILogger<PollutedLocationController> logger)
+    {
+        _logger = logger;
+    }
+
+    [HttpGet]
+    [Route("/[controller]/All")]
```
BenasB: Don't specify an absolute route here (starting with a `/`), instead of these two lines, just keep it relative to the controller's route:

```csharp
[HttpGet("All")]
```

And this action will be reachable at 

image

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
```diff
@@ -0,0 +1,55 @@
...
+[ApiController]
+[Route("/api/[controller]")]
+public class PollutedLocationController : ControllerBase
+{
+    private readonly ILogger<PollutedLocationController> _logger;
+
+    public PollutedLocationController(ILogger<PollutedLocationController> logger)
+    {
+        _logger = logger;
+    }
```
BenasB: If you're not using logger anywhere, you can safely remove it (and this constructor)

Apsitvarkom.Api/Apsitvarkom.Api.csproj
```diff
@@ -6,6 +6,10 @@
     <ImplicitUsings>enable</ImplicitUsings>
   </PropertyGroup>
 
+  <ItemGroup>
+    <Content Remove="package-lock.json" />
+  </ItemGroup>
+
```
BenasB: This does not delete the file itself, it just removes the file from the project that Visual Studio sees. If you go into file explorer, the file `package-lock.json` will still be there – delete the file and these lines in `Apsitvarkom.Api/Apsitvarkom.Api.csproj`

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
```diff
@@ -0,0 +1,55 @@
...
+            {
+                Id = "5be2354e-2500-4289-bbe2-66210592e17f",
+                Location = new LocationDTO
+                {
+                    Latitude = 12.1651,
+                    Longitude = 121.151,
+                },
+                Radius = 10,
+                Severity = "MODERATE",
+                Spotted  =  DateTime.Now.ToString("yyyy-MM-dd HH:mm:ss.ttt"),
```
BenasB: This timestamp format is not ISO8601 format, use `DateTime.UtcNow.ToString("o", CultureInfo.InvariantCulture)` to get this format: `"2022-09-22T09:00:29.4542617Z"`, like we specified in #4

BenasB: Of course use this for both DTOs you're creating

BenasB (changes requested):
Much more according to the plan now, just some small stuff left

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
BenasB: Refer to [this SO question](https://stackoverflow.com/questions/114983/given-a-datetime-object-how-do-i-get-an-iso-8601-date-in-string-format)

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
6ar8nas: My bad in this case. I suggested having it.

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
6ar8nas: To have more diversified data in DTOs I would have suggested we could have hard-coded one of the strings and not used two identical datetimes, although that's also good.

6ar8nas (approved):
Great job - looks good.

BenasB (approved):
LGTM, feel free to merge

PR #22, opened 2022-09-21
GetByIdAsync added to IPollutedLocationDTORepository

### LINKED ISSUE #19: Add GetByIdAsync method to IPollutedLocationDTORepository
## Description

Add `GetByIdAsync` method to [`IPollutedLocationDTORepository`](https://github.com/vu-vibedosa/apsitvarkom-be/blob/dev/Apsitvarkom.DataAccess/IPollutedLocationDTORepository.cs) interface and implement this method in the interface's implementation [`PollutedLocationsDTOFileRepository`](https://github.com/vu-vibedosa/apsitvarkom-be/blob/dev/Apsitvarkom.DataAccess/PollutedLocationsDTOFileRepository.cs).

## Steps to take

1. It might make sense to rename existing `GetAllPollutedLocationsAsync` to just `GetAllAsync` since the interface itself (I**PollutedLocationDTO**Repository) already indicates that this deals with `PollutedLocationDTO`s
2. Add `GetByIdAsync` method to the interface
3. Implement `GetByIdAsync` method in `PollutedLocationsDTOFileRepository`. This will most probably be just GetAll + Find. Think about what should happen if we can't find an entry with a given id: throw exception? return null? etc.
4. Write some unit tests checking the new method

## Acceptance criteria

1. `GetByIdAsync` added to the interface and implemented in the file repository implementation. 
2. Some unit tests written

6ar8nas:
## What was done

- GetByIdAsync method added to IPollutedLocationDTORepository
- Implemented GetByIdAsync in ~PollutedLocationDTOFileRepository
- Some file and method renaming for more consistency
- Unit tests written


Apsitvarkom.DataAccess/PollutedLocationDTOFileRepository.cs
```diff
@@ -25,27 +25,34 @@ private PollutedLocationsDTOFileRepository(Stream stream)
...
+    public async Task<IEnumerable<PollutedLocationDTO>> GetAllAsync()
     {
         var result = await JsonSerializer.DeserializeAsync<IEnumerable<PollutedLocationDTO>>(_stream, _options);
         return result ?? Enumerable.Empty<PollutedLocationDTO>();
     }
 
+    /// <inheritdoc />
+    public async Task<PollutedLocationDTO?> GetByIdAsync(string id)
+    {
+        var result = await GetAllAsync();
```
BenasB: NIT: `result` is a bit generic here, maybe a better name would be `allLocations` or something more specific?

Apsitvarkom.UnitTests/DataAccess/PollutedLocationDTOFileRepositoryTests.cs
```diff
@@ -103,43 +104,107 @@ public async Task GetAllPollutedLocationsDTO_JsonIncludesMultipleInstances_AllIn
...
+            $"{{\"id\":\"{id2}\"}}," +
+            $"{{\"id\":\"{id3}\"}}," +
+            $"{{\"id\":\"{id2}\"}}" +
+            "]";
+        using var dataManager = PollutedLocationDTOFileRepository.FromContent(jsonString);
+
+        var requestId = id2;
+
+        Assert.That(Assert.ThrowsAsync<InvalidOperationException>(async () => await dataManager.GetByIdAsync(requestId))!.Message, 
+            Is.EqualTo("Sequence contains more than one matching element"));
```
BenasB: Why do you need to check with a hard coded string? `Assert.ThrowsAsync` alone will do just what you need – assert if it throws. I don't see a reason to check the exception message

BenasB (changes requested):
One nitpick and the test assert would be a good fix

BenasB (approved):
LGTM, feel free to merge

PR #25, opened 2022-09-23
Distance to location implementation

### LINKED ISSUE #23: DistanceTo extension method for Location class
## Description

One of the requirements for Lab 1 is
> 4. Extension method usage;

It would be good (maybe we'll use it in the future) to have a `DistanceTo(Location otherLocation)` method to calculate the distance between two locations/coordinates. We could just create this method in the [`Location class`](https://github.com/vu-vibedosa/apsitvarkom-be/blob/dev/Apsitvarkom.Models/Location.cs), but we could also fulfill the requirement of using an extension method.

## Steps to take

1. Create an extension method for `Location` class called `DistanceTo` that will take in another `Location` and calculate the distance between them in meters. You'll have to do some research yourself on how to calculate it mathematically, but there should be resources available such as [this SO answer](https://stackoverflow.com/a/60809937).
2. Write unit tests for this extension method

P.S. I see that there are packages such as [Geolocation](https://www.nuget.org/packages/Geolocation/) or [Geolocation.NetStandard](https://www.nuget.org/packages/Geolocation.NetStandard/), but they might be overkill for our use case.

## Acceptance criteria

1. Extension method `DistanceTo` is created for class `Location`
2. Some unit tests written

6ar8nas:
## What was done

- Made an extension method for calculating distance between two locations (coordinates).
- Wrote unit tests for this extension.

### DIFF
--- a/Apsitvarkom.Api/Extensions.cs
+++ b/Apsitvarkom.Api/Extensions.cs
@@ -1,0 +1,23 @@
+﻿using Apsitvarkom.Models;
+
+namespace Apsitvarkom.Api;
+
+/// <summary>
+/// Class for extensions used in this solution.
+/// </summary>
+public static class Extensions
+{
+    /// <summary>Distance from source to destination coordinates using the Haversine formula.</summary>
+    /// <param name="source">Coordinate to calculate distance from.</param>
+    /// <param name="destination">Coordinate to calculate distance to.</param>
+    /// <returns>Distance from source to destination coordinates in meters.</returns>
+    public static double DistanceTo(this Location source, Location destination)
+    {
+        const double earthRadius = 6364050.0; // Calculated to Vilnius's latitude on ground level
+        var deltaLat = (destination.Latitude - source.Latitude) * (Math.PI / 180.0);
+        var deltaLon = (destination.Longitude - source.Longitude) * (Math.PI / 180.0);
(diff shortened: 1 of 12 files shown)


BenasB (changes requested):
Method implementation and tests looking solid, although I don't understand why the extension method is located in `Apsitvarkom.Api` project – it has nothing to do with our API? I'd place it in `Apsitvarkom.Models` project since this method only concerns `Location`. Moreover, you don't necessarily have to put all of your extension methods in one file – in fact that wouldn't make much sense as the project grows, since it could get huge. 

What about moving and renaming `Apsitvarkom.Api/Extensions.cs` to `Apsitvarkom.Models/ModelExtensions.cs` ? We would then have all model related extensions in `Apsitvarkom.Models/ModelExtensions.cs`  and if there's needed extensions for the Api project, we'll create a new file there then.

Unless you have some arguments why we should keep it as is

6ar8nas:
Good insights.

> What about moving and renaming Apsitvarkom.Api/Extensions.cs to Apsitvarkom.Models/ModelExtensions.cs ?

Didn't comply to this suggestion to the fullest though. As each extensions class is in its own project, I don't think an explicit naming for '<ProjectName>Extensions' are needed and just `Extensions` is enough.

BenasB (approved):
LGTM, feel free to merge

PR #36, opened 2022-09-27
Write Unit tests for PollutedLocationController using Moq

### LINKED ISSUE #33: Write Unit tests for PollutedLocationController using Moq
## Description

Write Unit tests for all methods of PollutedLocationController using [`Moq`](https://github.com/moq/moq4).

Originally requested by @BenasB in [this](https://github.com/vu-vibedosa/apsitvarkom-be/pull/31#pullrequestreview-1120711500) comment in #31.
      
## Steps to take

1. Install `Moq` NuGet package.
2. Cover PollutedLocationController methods with Unit tests mocking repository.
3. Make sure you cover all possible cases of HTTPResponses (`200`, `404`, `500`).

## Acceptance criteria

Tests written for all PollutedLocationController methods.

6ar8nas:
## What was done

1. Moq NuGet package installed.
2. PollutedLocationController methods covered with Unit tests mocking repository.
3. All possible controller created cases of HTTPResponses covered.

### DIFF
--- a/Apsitvarkom.UnitTests/Api/Controllers/PollutedLocationControllerTests.cs
+++ b/Apsitvarkom.UnitTests/Api/Controllers/PollutedLocationControllerTests.cs
@@ -1,0 +1,117 @@
+﻿using Apsitvarkom.Api.Controllers;
+using Apsitvarkom.DataAccess;
+using Apsitvarkom.Models.DTO;
+using Microsoft.AspNetCore.Mvc;
+using Moq;
+
+namespace Apsitvarkom.UnitTests.Api.Controllers;
+
+public class PollutedLocationControllerTests
+{
+    private Mock<IPollutedLocationDTORepository> m_repository;
+    private PollutedLocationController m_controller;
+
+    [SetUp]
+    public void SetUp()
+    {
+        m_repository = new Mock<IPollutedLocationDTORepository>();
+        m_controller = new PollutedLocationController(m_repository.Object);
+    }
+
+    [Test]
+    public void Constructor_HappyPath_IsSuccess() => Assert.That(new PollutedLocationController(m_repository.Object), Is.Not.Null);
+
+    [Test]
+    public async Task GetAll_RepositoryReturnsPollutedLocationDTOs_OKActionResultReturned()
+    {
+        var instances = GetPollutedLocationDTOs();
+        m_repository.Setup(self => self.GetAllAsync()).ReturnsAsync(instances);
+
+        var actionResult = await m_controller.GetAll();
+
+        var result = actionResult.Result as OkObjectResult;
+
+        Assert.That(result, Is.Not.Null);
+        Assert.That(result!.StatusCode, Is.EqualTo(200));
+        Assert.That(result.Value, Is.EqualTo(instances));
+    }
+
+    [Test]
+    public async Task GetById_RepositoryReturnsPollutedLocationDTO_OKActionResultReturned()
+    {
+        var instanceId = Guid.NewGuid().ToString();
+        var instance = new PollutedLocationDTO
+        {
+            Id = instanceId,
+            Location = new LocationDTO
+            {
+                Latitude = 111.11111,
+                Longitude = 11.11111
+            },
+            Radius = 11,
+            Severity = "Low",
+            Spotted = "2023-11-23T21:12:14Z",
+            Progress = 11,
+            Notes = "11111"
+        };
+        m_repository.Setup(self => self.GetByIdAsync(instanceId)).ReturnsAsync(instance);
+
+        var actionResult = await m_controller.GetById(instanceId);
+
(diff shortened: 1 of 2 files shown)


Apsitvarkom.UnitTests/Api/Controllers/PollutedLocationControllerTests.cs
```diff
@@ -0,0 +1,117 @@
...
+                },
+                Radius = 11,
+                Severity = "Low",
+                Spotted = "2023-11-23T21:12:14Z",
+                Progress = 11,
+                Notes = "11111"
+            }
+        };
+        return instances;
+    }
```
BenasB: This whole method can become a property, see https://onecompiler.com/csharp/3yhcy6mdq

BenasB: or at least an expression bodied method

Apsitvarkom.UnitTests/Api/Controllers/PollutedLocationControllerTests.cs
```diff
@@ -0,0 +1,117 @@
...
+    {
+        var instanceId = Guid.NewGuid().ToString();
+        m_repository.Setup(self => self.GetByIdAsync(instanceId)).ReturnsAsync((PollutedLocationDTO?)null);
+
+        var actionResult = await m_controller.GetById(instanceId);
+
+        var result = actionResult.Result as NotFoundObjectResult;
+
+        Assert.That(result, Is.Not.Null);
+        Assert.That(result!.StatusCode, Is.EqualTo(404));
```
BenasB: what's that magic number? :D

Why not use something like we used [here](https://github.com/vu-vibedosa/apsitvarkom-be/blob/dev/Apsitvarkom.Api/Controllers/PollutedLocationController.cs#L18)?

BenasB: And everywhere else of course

Apsitvarkom.UnitTests/Api/Controllers/PollutedLocationControllerTests.cs
```diff
@@ -0,0 +1,117 @@
...
+        Assert.That(result, Is.Not.Null);
+        Assert.That(result!.StatusCode, Is.EqualTo(200));
+        Assert.That(result.Value, Is.EqualTo(instances));
+    }
+
+    [Test]
+    public async Task GetById_RepositoryReturnsPollutedLocationDTO_OKActionResultReturned()
+    {
+        var instanceId = Guid.NewGuid().ToString();
+        var instance = new PollutedLocationDTO
```
BenasB: Any reasons why create a new object here and not for example `GetPollutedLocationDTOs[0]`?

BenasB: or `.First()`

BenasB (changes requested):
Couple of comments

Apsitvarkom.UnitTests/Api/Controllers/PollutedLocationControllerTests.cs
```diff
@@ -0,0 +1,99 @@
...
+using Moq;
+
+namespace Apsitvarkom.UnitTests.Api.Controllers;
+
+public class PollutedLocationControllerTests
+{
+    private PollutedLocationController m_controller;
+    private Mock<IPollutedLocationDTORepository> m_repository;
+
+    private readonly IEnumerable<PollutedLocationDTO> PollutedLocationDTOs = new List<PollutedLocationDTO>
```
BenasB, line 15: This is just a field now and no matter how many times you get the value of this field, the `Id`s will be constant – field value is assigned when this class (`PollutedLocationControllerTests`) is created. If you wanted the `Id` to be a different value each time you retrieve this `PollutedLocationDTOs`, you would need to turn this **field** into a **property with a getter**. You can use the short syntax by simply changing the assign `=` symbol into the expression symbol `=>`

BenasB (approved):

Apsitvarkom.UnitTests/Api/Controllers/PollutedLocationControllerTests.cs
BenasB: This is not something that you need to fix (since you don't really care about the `Id` field), just a note on the behavior

6ar8nas: Cool, will have that in mind later. Thanks. I'm resolving this comment.

EDIT: found a [place](https://github.com/vu-vibedosa/apsitvarkom-be/blob/dev/Apsitvarkom.UnitTests/Models/Validation/PollutedLocationValidationTests.cs#L9) where we have already used it for future references.

PR #42, opened 2022-10-04
Regex validation

### LINKED ISSUE #37: Regex check sourcePath in PollutedLocationDTOFileRepository
## Description

There is a requirement for Lab 1:
> 7. Regex usage;

We have a [`FromFile`](https://github.com/vu-vibedosa/apsitvarkom-be/blob/dev/Apsitvarkom.DataAccess/PollutedLocationDTOFileRepository.cs#L43) method in `PollutedLocationDTOFileRepository`. We could add regex checking/validating the `sourcePath` string here. I understand that `File.OpenRead` will throw if the file path is bad, but I suggest this just for the sake of completing the requirement.

## Steps to take

1. Determine what criteria should be met for the `sourcePath` string. No spaces? Ends with file extension `.json`? etc.
2. Create the regex, https://regex101.com/ could help
3. Implement the regex check in `FromFile` method, throw an exception if it does not match the regex.
4. Unit test showing the functionality may be nice

## Acceptance criteria

1. `FromFile` does regex matching/validation on the `sourcePath` argument and throws if it fails.

BaVilius:
Implemented a method that uses Regex to validate whether file has a `.json` extension

### DIFF
--- a/Apsitvarkom.DataAccess/PollutedLocationDTOFileRepository.cs
+++ b/Apsitvarkom.DataAccess/PollutedLocationDTOFileRepository.cs
@@ -1,5 +1,4 @@
-﻿using System.Net.Http.Headers;
-using System.Text;
+﻿using System.Text;
 using System.Text.Json;
 using System.Text.RegularExpressions;
 using Apsitvarkom.Models.DTO;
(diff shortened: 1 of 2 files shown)


6ar8nas (approved):
looks good. thanks

Issue #44, opened 2022-10-05
Inherit from Location in PollutedLocation


BenasB:
## Description

There is a requirement: 
> 8. Widening and narrowing type conversions;

As discussed in lecture on 2022-10-04, lets achieve this by using inheritance instead of composition ([inheritance vs composition](https://www.digitalocean.com/community/tutorials/composition-vs-inheritance)) for `Location` and `PollutedLocation` classes. The same needs to be done for DTOs as well.

## Steps to take

1. In `PollutedLocation`, instead of having `Location` [as a property](https://github.com/vu-vibedosa/apsitvarkom-be/blob/dev/Apsitvarkom.Models/PollutedLocation.cs#L15) inherit from it:
```csharp
public class PollutedLocation : Location
{
    // No `public Location Location { get; set; } = null!;` here

    // Other properties
}
```
2. In [`Location`](https://github.com/vu-vibedosa/apsitvarkom-be/blob/dev/Apsitvarkom.Models/Location.cs) instead of having `Latitude` and `Longitude` directly, move them to a `struct Coordinates` (the `Coordinates` struct itself can exist in the same file or you can move it out to its own file):
```csharp
public class Location
{
    public Coordinates Coordinates { get; set; }
}
```
**This will let us keep similar json structure, but the `location` key will be replaced by `coordinates` key**
3. Do the same 
4. Adjust validators, mappers and unit tests if needed 

## Acceptance criteria

1. `PollutedLocation` is derived from `Location`
2. `PollutedLocationDTO` is derived from `LocationDTO`
3. `Location` and `LocationDTO` classes have one property `Coordinates` (of types `Coordinates` and `CoordinatesDTO` respectively)
4. Validators, mappers and unit tests working as expected, no errors

PR #47, opened 2022-10-05
Add GetAllAsync(Location inRelationTo) overload to IPollutedLocationDTORepository

### LINKED ISSUE #38: Add GetAllAsync(Location inRelationTo) overload to IPollutedLocationDTORepository
## Description

There is a requirement for Lab 1:

> 9. LINQ to Objects usage (methods and queries);

And we don't have a LINQ query (words syntax) yet (we only use LINQ methods). To make better use of the query syntax, we need to have some sort of a filtering/ordering need. I propose to create an overload to [`GetAllAsync`](https://github.com/vu-vibedosa/apsitvarkom-be/blob/dev/Apsitvarkom.DataAccess/IPollutedLocationDTORepository.cs#L12) which would take in a [`Location`](https://github.com/vu-vibedosa/apsitvarkom-be/blob/dev/Apsitvarkom.Models/Location.cs) as an argument and then return all of the polluted locations **ordered by their distance from the given location** (ascending, from the smallest to the largest).

## Steps to take

1. Define the new method overload in the [repository interface](https://github.com/vu-vibedosa/apsitvarkom-be/blob/dev/Apsitvarkom.DataAccess/IPollutedLocationDTORepository.cs).
2. Implement the new method in the [file repository implementation](https://github.com/vu-vibedosa/apsitvarkom-be/blob/dev/Apsitvarkom.DataAccess/PollutedLocationDTOFileRepository.cs).
    1. To calculate the actual distance use the extension method we created in #23
    2. Use the [**query (word) LINQ syntax**](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/linq/basic-linq-query-operations) for ordering/sorting.
3. Write unit test(s) for the new method

## Acceptance criteria

1. `IPollutedLocationDTORepository` has a new method
2. The new method is implemented in `PollutedLocationDTOFileRepository`
3. Unit test(s) written

6ar8nas:
## What was done

- Added an overload for GetAllSync that allows ordering by distance.
- Wrote tests to test this repository behavior.

### DIFF
--- a/Apsitvarkom.Api/Program.cs
+++ b/Apsitvarkom.Api/Program.cs
@@ -1,13 +1,18 @@
 using Apsitvarkom.DataAccess;
 using Apsitvarkom.Models.DTO;
 using Apsitvarkom.Models.Mapping;
+using AutoMapper;
 using FluentValidation;
 
 var builder = WebApplication.CreateBuilder(args);
 
 // Add services to the container.
 
-builder.Services.AddScoped<IPollutedLocationDTORepository>(_ => PollutedLocationDTOFileRepository.FromFile("PollutedLocationsMock.json"));
+builder.Services.AddScoped<IPollutedLocationDTORepository>(_ => PollutedLocationDTOFileRepository.FromFile(new MapperConfiguration(cfg =>
+{
+    cfg.AddProfile<PollutedLocationProfile>();
+}).CreateMapper(), "PollutedLocationsMock.json"));
+
 builder.Services.AddControllers();
 // Learn more about configuring Swagger/OpenAPI at https://aka.ms/aspnetcore/swashbuckle
 builder.Services.AddEndpointsApiExplorer();
--- a/Apsitvarkom.Api/Program.cs
+++ b/Apsitvarkom.Api/Program.cs
@@ -8,11 +8,6 @@ var builder = WebApplication.CreateBuilder(args);
 
 // Add services to the container.
 
-builder.Services.AddScoped<IPollutedLocationDTORepository>(_ => PollutedLocationDTOFileRepository.FromFile(new MapperConfiguration(cfg =>
-{
-    cfg.AddProfile<PollutedLocationProfile>();
-}).CreateMapper(), "PollutedLocationsMock.json"));
-
 builder.Services.AddControllers();
 // Learn more about configuring Swagger/OpenAPI at https://aka.ms/aspnetcore/swashbuckle
 builder.Services.AddEndpointsApiExplorer();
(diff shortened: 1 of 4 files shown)


Apsitvarkom.DataAccess/PollutedLocationDTOFileRepository.cs
```diff
@@ -23,6 +27,7 @@ private PollutedLocationDTOFileRepository(Stream stream)
             AllowTrailingCommas = true,
             PropertyNameCaseInsensitive = true
         };
+        _mapper = new MapperConfiguration(cfg => cfg.AddProfile<PollutedLocationProfile>()).CreateMapper();
```
6ar8nas, line 30: didn't find a way to use DI here as static factory is necessary to support both ways of the repository. maybe I am missing something?

Apsitvarkom.Api/Program.cs
```diff
@@ -1,13 +1,18 @@
...
 var builder = WebApplication.CreateBuilder(args);
 
 // Add services to the container.
 
-builder.Services.AddScoped<IPollutedLocationDTORepository>(_ => PollutedLocationDTOFileRepository.FromFile("PollutedLocationsMock.json"));
+builder.Services.AddScoped<IPollutedLocationDTORepository>(_ => PollutedLocationDTOFileRepository.FromFile(new MapperConfiguration(cfg =>
+{
+    cfg.AddProfile<PollutedLocationProfile>();
+}).CreateMapper(), "PollutedLocationsMock.json"));
+
```
BenasB:
```csharp
builder.Services.AddScoped<IPollutedLocationDTORepository>(serviceProvider =>
{
    var mapper = serviceProvider.GetRequiredService<IMapper>();
    return PollutedLocationDTOFileRepository.FromFile(mapper, "PollutedLocationsMock.json");
});
```

BenasB: `IMapper` will be automatically added to dependency injection container by [`AddAutoMapper` method](https://github.com/vu-vibedosa/apsitvarkom-be/blob/dev/Apsitvarkom.Api/Program.cs#L17).

BenasB (changes requested):
LGTM apart from the DI part

Apsitvarkom.Api/Program.cs
6ar8nas: Thanks, this makes sense. I was just confused to what would happen if in the future we would have more AutoMapper profiles - which one would be chosen? - That's the reason of me sticking with this implementation, that was way off of seeming to be ideal, but I was clear would work right. 

Fixed it. I'm also not sure if that was necessary but I also moved the statement below AddAutoMapper. After testing it, it seemed to work just fine both ways - is there any prefered one? Let me know if I should revert this.

BenasB: > if in the future we would have more AutoMapper profiles - which one would be chosen?

As I understand it, profiles are only used to group mappings. When you register automapper (call `AddAutoMapper`) to DI, it picks up all profiles and adds an instance of `IMapper` to the DI container – this instance knows all mappings (combined from all profiles). What would happen if you were to define two *mappings* (does not matter if they're in different *profiles* or not) from `Type1` to `Type2` I can't tell, I expect it to throw an error because it won't know which one to choose – but either way you should probably not end up in this situation.

> is there any prefered one?

The order should not matter – the services you register to DI get evaluated/created later (not at the time where you declare them). There's no preferred way AFAIK

BenasB (approved):
LGTM, responded to comments, feel free to merge. This will cause merge conflicts with #44 but whoever is doing that task will need to deal with that

PR #55, opened 2022-10-11
Type conversions

### LINKED ISSUE #44: Inherit from Location in PollutedLocation

BaVilius:
There is still a problem where I can't figure out how to set LocationDTOValidator() for validating Coordinates, hence the tests failing

### DIFF
--- a/Apsitvarkom.DataAccess/Apsitvarkom.DataAccess.csproj
+++ b/Apsitvarkom.DataAccess/Apsitvarkom.DataAccess.csproj
@@ -10,4 +10,8 @@
     <ProjectReference Include="..\Apsitvarkom.Models\Apsitvarkom.Models.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
+  </ItemGroup>
+
 </Project>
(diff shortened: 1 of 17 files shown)


BenasB:
## What was changed

- Use Newtonsoft.Json instead of System.Text.Json, because [it does not support deserializing polymorphic objects ](https://stackoverflow.com/questions/58074304/is-polymorphic-deserialization-possible-in-system-text-json). But Newtonsoft.Json does not have asynchronous deserialization, so I had to do some fake async
- Adjust tests
- Adjust mappers
- Adjust validators

BenasB:
We're breaking the json schema! Front end will need updating

BenasB (approved):

PR #56, opened 2022-10-11
Api use Newtonsoft.Json

BenasB:
## What was done

- Make Api use Newtonsoft.Json instead of System.Text.Json


6ar8nas (approved):
looks good

PR #57, opened 2022-10-11
Quick Fix: Variables can be created using var

6ar8nas:
## What was done

Changed initialization of variables to use var instead of specific class.


BenasB (approved):
LGTM

PR #60, opened 2022-10-12
QuickFix: Collaborative code touch-up during the lecture

6ar8nas:
## What was done

- Fulfilled the widening-narrowing requirement.
- Fulfilled the named argument usage requirement.

### DIFF
--- a/Apsitvarkom.DataAccess/PollutedLocationDTOFileRepository.cs
+++ b/Apsitvarkom.DataAccess/PollutedLocationDTOFileRepository.cs
@@ -46,7 +46,7 @@ public class PollutedLocationDTOFileRepository : IPollutedLocationDTORepository,
     public async Task<IEnumerable<PollutedLocationDTO>> GetAllAsync(Location inRelationTo)
     {
         return from pollutedLocation in await GetAllAsync()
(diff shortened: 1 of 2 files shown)


Apsitvarkom.DataAccess/PollutedLocationDTOFileRepository.cs
```diff
@@ -46,7 +46,7 @@ public Task<IEnumerable<PollutedLocationDTO>> GetAllAsync()
     public async Task<IEnumerable<PollutedLocationDTO>> GetAllAsync(Location inRelationTo)
     {
         return from pollutedLocation in await GetAllAsync()
-               orderby inRelationTo.DistanceTo(_mapper.Map<Location>(pollutedLocation))
+               orderby inRelationTo.DistanceTo(_mapper.Map<PollutedLocation>(pollutedLocation))
```
BenasB: Variable name `pollutedLocation` is confusing here. Right now it seems like it's a `PollutedLocation` that gets mapped into the same `PollutedLocation`. It's really a DTO though, so maybe a better name would be `pollutedLocationDTO` or just `dto` etc.

BenasB (approved):
LGTM apart from one comment

PR #61, opened 2022-10-12
Implement a generic interface for our used repository

### LINKED ISSUE #58: Implement a generic interface for our used repository
## Description

For our requirement we have to use generics. Let's change our repository interface to be generic. 

## Steps to take

1. Change the interface `IPollutedLocationDTORepository` to be [generic](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/generics/generic-interfaces). You can name it `IEntityDTORepository`.
2. Adjust the existent implementation of `PollutedLocationDTOFileRepository` to implement the new interface.
3. Adapt unit tests if necessary.

## Acceptance criteria

Usage of a generic interface to complete the requirement.

6ar8nas:
## What was done

- Changed the previous interface `IPollutedLocationDTORepository` to be generic.
- Adjusted the existent implementation of PollutedLocationDTOFileRepository to implement the newly created interface.


BaVilius (approved):
Approved

Apsitvarkom.DataAccess/IEntityDTORepository.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace Apsitvarkom.DataAccess;
+
+/// <summary>
+/// Interface for <see cref="T" /> class data handling.
+/// </summary>
+public interface IEntityDTORepository<T> where T: class
```
BenasB: Any specific reason why `Entity` here in the name? I would suggest not adding one more word to our code just to not confuse people. Usually more words for the same thing = more confusion.

What about `IModelDTORepository`, `IDTORepository` or just `IRepository` ?

Apsitvarkom.DataAccess/IEntityDTORepository.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    /// <summary>Gets all <see cref="T" /> records from the data source.</summary>
+    /// <returns><see cref="Enumerable" /> of <see cref="T" /> instances.</returns>
+    public Task<IEnumerable<T>> GetAllAsync();
+
+    /// <summary>Gets a single <see cref="T" /> record from the data source by id.</summary>
+    /// <param name="id">Identifier of the requested record.</param>
+    /// <returns>Null if the instance was not found. Otherwise a <see cref="T" /> instance with matching identifier.</returns>
+    public Task<T?> GetByIdAsync(string id);
```
BenasB: What if type `T` does not have an `Id` property, or that property's type is not `string`? You have to keep these things in mind when creating a *generic* interface (you either limit the generic type so that you will be certain that the generic type will have the fields you need or you generalize the methods more).

Maybe we could go for `public Task<T?> GetByPropertyAsync(Predicate<T> propertyCondition);`

That could be used something like this:
```csharp
const string idToSearch = "this-is-an-id";
var dto = _repository.GetByPropertyAsync(x => x.Id == myId);

const int amountToSearch = 42;
var otherDto = _otherTypeRepository.GetByPropertyAsync(x => x.Amount == amountToSearch);
```

Apsitvarkom.DataAccess/ILocationDTORepository.cs
```diff
@@ -0,0 +1,15 @@
+﻿using Apsitvarkom.Models;
+using Apsitvarkom.Models.DTO;
+
+namespace Apsitvarkom.DataAccess;
+
+/// <summary>
+/// Interface for <see cref="T" /> class data handling, where <see cref="T"/> derives from <see cref="LocationDTO"/>.
+/// </summary>
+public interface ILocationDTORepository<T> : IEntityDTORepository<T> where T : LocationDTO
```
BenasB: I wonder if we're planning on having multiple types derived from `LocationDTO`. We now only have `PollutedLocationDTO` and I didn't think we'll be adding more types derived from `LocationDTO`, so my initial thought was to have this interface non generic:

`public interface IPollutedLocationDTORepository : IEntityDTORepository<PollutedLocationDTO>`

Just because you *could* have it generic, doesn't mean that you *need* it to have generic. Although we can keep it as it is right now if you think we could be adding more types deriving from `LocationDTO` in the future.

BenasB (changes requested):
Design questions

Apsitvarkom.DataAccess/ILocationDTORepository.cs
6ar8nas: I wasn't sure on how we would handle those `Event` type classes - whether we would derive it from PollutedLocationDTO, have the related PollutedLocationDTO as a property inside of it or have some separate class without much direct relation to the PollutedLocationDTO (hence probably needing to derive it from LocationDTO). Maybe let's keep it as is until we figure this one out and could change it if needed afterwards?

Apsitvarkom.DataAccess/IEntityDTORepository.cs
6ar8nas: yuup, makes sense. Just I believe I had to use not predicate, but func there.

BenasB: I would prefer a `Predicate<T>` since it's achieves the same thing as `Func<T, bool>` (represents a function that always returns `bool`), but a quick google search [leads to this](https://stackoverflow.com/a/665513), so we can keep it just for the sake of LINQ's consistency

Apsitvarkom.DataAccess/PollutedLocationDTOFileRepository.cs
```diff
@@ -51,10 +51,10 @@ orderby inRelationTo.DistanceTo(_mapper.Map<PollutedLocation>(pollutedLocationDT
     }
 
     /// <inheritdoc />
-    public async Task<PollutedLocationDTO?> GetByIdAsync(string id)
+    public async Task<PollutedLocationDTO?> GetByPropertyAsync(Func<PollutedLocationDTO, bool> propertyCondition)
     {
         var allLocations = await GetAllAsync();
-        return allLocations.SingleOrDefault(loc => loc.Id == id);
+        return allLocations.SingleOrDefault(propertyCondition);
```
BenasB: Is this still `SingleOrDefault` since we can "search" based on different properties? Maybe `FirstOrDefault`? Just a thought

BenasB (approved):
Read over the comments I added and if you think there's no need to change anything, resolve those conversations and feel free to merge

Apsitvarkom.DataAccess/ILocationDTORepository.cs
BenasB: Sure, we can keep it as is for now and change it if needed. 

> I wasn't sure on how we would handle those Event type classes

Composition probably (that is the "have the related PollutedLocationDTO as a property inside of it" option). In a database we will probably store a foreign key to the polluted location ID.

PR #68, opened 2022-10-23
reverse geocoder api call

### LINKED ISSUE #66: Implement Geocoder class
## Description

As we've discussed earlier, we want to use Google Maps Platform [Geocoding API](https://developers.google.com/maps/documentation/geocoding) to get "Titles" for our locations/coordinates (that process is called reverse geocoding). Let's start implementing this by creating a `Geocoder` class that will **handle making request to Google's Geocoding API via the .NET's `HttpClient`**. We could use a library such as https://github.com/vivet/GoogleApi or https://github.com/maximn/google-maps but these seem overkill for our case (and not that much maintained) IMO.

## Steps to take

1. Create class `Geocoder` in `DataAccess` project. In this class you will be making HTTP requests to Google's [Geocoding API](https://developers.google.com/maps/documentation/geocoding).
2. We are interested in reverse geocoding process, so study [the docs about it](https://developers.google.com/maps/documentation/geocoding/requests-reverse-geocoding) – this page will probably have everything you need from their API. Main things to think about:
    1. What endpoint/URL to call
    2. How to make the request (URL query parameters)
    3. What response (what JSON object/fields) will you get and what do you need from it
4. To make the actual HTTP calls use in built class `HttpClient`. Here's the [official MS documentation](https://learn.microsoft.com/en-us/dotnet/csharp/tutorials/console-webapiclient) on how to use it.
5. Create an interface `IGeocoder` that has one method `string ReverseGeocode(Coordinates)` – this is an abstraction of the process: take in coordinates and return a string/title/address from them.
6. Implement the interface `IGeocoder` in `Geocoder` class, namely the `ReverseGeocode` method where you call the Google's Geocoding API .

Note: As you can see from the Geocoding API docs, you need an API key to make the requests. This API key is located in a Google sheet where all the team members were added. **Do not put that API key in source control**. Use [Secret Manager to store it locally](https://learn.microsoft.com/en-us/aspnet/core/security/app-secrets?view=aspnetcore-6.0&tabs=windows).

## Acceptance criteria

`Geocoder` class implemented that inherits `IGeocoder` and has one method `ReverseGeocode` that takes in `Coordinates` type and returns a `string` representing the title/address for the given location.

XxDoMaSxX:
## What was done

-class Geocoder that converts coordinates to location title
-iGeocoder interface with implemented function


Apsitvarkom.DataAccess/IGeocoder.cs
```diff
@@ -0,0 +1,14 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Apsitvarkom.DataAccess
+{
+    public interface IGeocoder
+    {
+        public Task<string> ReverseGeocode(Coordinates cords);
```
BenasB: I think you don't even need `public` here, a method declaration in an interface expects the method to be `public` in the implementing class by default

Apsitvarkom.UnitTests/DataAccess/PollutedLocationDTOFileRepositoryTests.cs
```diff
@@ -270,5 +270,21 @@ public async Task GetByPropertyAsync_SeveralInstanceWithRequestedIdFound_FirstMa
...
+        {
+            Longitude = 55.7329098,
+            Latitude = 24.3557527,
+        };
+       
+        var geocoderClass = new Geocoder();
+        var result = geocoderClass.ReverseGeocode(coordinates);
+
+
+        Assert.AreEqual("test", result.Result);
```
BenasB: Use `Assert.That` like in the GitHub check annotation

> Consider using the constraint model, Assert.That(actual, Is.EqualTo(expected)), instead of the classic model, Assert.AreEqual(expected, actual)

Apsitvarkom.DataAccess/Geocoder.cs
```diff
@@ -0,0 +1,35 @@
...
+    public class Geocoder : IGeocoder
+    {
+        private static readonly HttpClient client = new HttpClient();
+
+        public static string geocoderApiKey;
+
+        public async Task<string> ReverseGeocode(Coordinates cords)
+        {
+            var apiResponse = await MakeReverseGeocodeApiCall(cords);
+            dynamic dynamicObject = JsonConvert.DeserializeObject<dynamic>(apiResponse)!;
```
BenasB: Instead of a `dynamic` you should define a class `ReverseGeocodingApiResponse` (or something similar) and describe the type you are expecting to get back from the API (only the fields you need, other fields will probably just get ignored).

I did not test this, but something like this should work:

```csharp
private class ReverseGeocodingApiResponse {
    public IEnumerable<ReverseGeocodingApiResult> Results;
}

private class ReverseGeocodingApiResult {
    public string FormattedAddress;
}
```

you will need to check how the keys/names are handled (in c# we have `FormattedAddress` but API actually send us back `formatted_address`)

BenasB: We want to do it like this so we have static type checking - we know what to expect when writing code (in contrast `dynamic` does not inform us about the data that is inside)

Apsitvarkom.DataAccess/Geocoder.cs
```diff
@@ -0,0 +1,35 @@
...
+            string locationTitle = dynamicObject.results[0].formatted_address;
+            return locationTitle;
+        }
+
+        private static async Task<string> MakeReverseGeocodeApiCall(Coordinates cordinates)
+        {
+            var apiCall = client.GetStringAsync("https://maps.googleapis.com/maps/api/geocode/json?latlng=54.687203,25.241313&language=lt&result_type=political&key=[ENTER_API_KEY_THERE]");
```
BenasB: You probably want to substitute real coordinates values in query parameter `latlng` instead of hard coding`54.687203,25.241313`

BenasB (changes requested):
For passing in the API key to the `Geocoder`, you can do it like this:

Firstly, do not use a static field in the `Geocoder` class to store it, you can remove that static field and these two lines from `Program.cs`

```csharp
var geocoderApiKey = app.MapGet("/", () => apiKey);
Geocoder.geocoderApiKey = geocoderApiKey.ToString();
```

Using a static field like this is bad, because as you can see it can be assigned/retrieved from anywhere in our code - that is not a good practice, we should instead limit the access to data only for the parts that actually use it.

Secondly, as a simple solution, you should just create a constructor for `Geocoder` that takes in a `string` for our API key and just save it in a `private readonly string` field. Something like:

```csharp
public class Geocoder : IGeocoder
{
    // other fields
    private readonly string _apiKey;

    public Geocoder (string apiKey) {
        _apiKey = apiKey;
    }

    // Other methods where you use _apiKey variable
}
```

And in `Program.cs` you can retrieve the API key from `IConfigration` like this: 
`var mapsApiKey = builder.Configuration.GetValue<string>("Maps:GoogleMapsApiKey")` (of course you need to set it in your secrets first) and then register `Geocoder` to dependency injection container: `builder.Services.AddSingleton<IGeocoder>(_ => new Geocoder(mapsApiKey))`.

As for unit tests, you would need to use `Moq` to mock the httpClient, but that will take additional knowledge and time, so we can do it in the next task. For local manual testing you can create an action on a new controller (for example `GeocodingController`) and constructor inject the `IGeocoder` instance (the one you registered with `AddSingleton` previously):

Oversimplified controller:
```csharp
public class GeocodingControler {
    private readonly IGeocoder _geocoder;

    public GeocodingController(IGeocoder geocoder){
        // `geocoder` will automatically be in the constructor because you registered it with AddSingleton in Program.cs
        _geocoder = geocoder;
    }

    // Action using _geocoder
}
```

This way you can manually test the geocoder in swagger

6ar8nas:
Pay attention to the github checks - it has some accurate comments. There are also some places where you use two new lines, which is usually not done in enterprise level (at least from my experience).

Apsitvarkom.UnitTests/DataAccess/PollutedLocationDTOFileRepositoryTests.cs
```diff
@@ -270,5 +270,21 @@ public async Task GetByPropertyAsync_SeveralInstanceWithRequestedIdFound_FirstMa
 
         Assert.That(instance!.Radius, Is.EqualTo(2));
     }
+
+    [Test]
+    public async Task tempTest()
```
6ar8nas, line 275: you don't need `async Task` here as is in the test above, as the method ReverseGeocode is not async. considering using `void` instead

Apsitvarkom.DataAccess/IGeocoder.cs
6ar8nas: We could leave it as is though, as all our written interfaces have the access modifier public.
The other option could be also editing already existing interfaces to match it?

Also, I believe it defaults to internal and not public, at least so is written [here](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/access-modifiers#other-types)

Apsitvarkom.DataAccess/IGeocoder.cs
```diff
@@ -0,0 +1,14 @@
+﻿using Apsitvarkom.Models;
+using System;
+using System.Collections.Generic;
+using System.Linq;
```
6ar8nas: mostly all these `using` directives are not used. delete the not necessary ones. Check for the same issue in Geocoder.cs too

Apsitvarkom.DataAccess/IGeocoder.cs
BenasB: I'm talking about the `public` in `public Task<string> ReverseGeocode(Coordinates cords)`, sorry if that wasn't clear. In the page you linked it says:

> Interface members are public by default because the purpose of an interface is to enable other types to access a class or struct.

Apsitvarkom.UnitTests/DataAccess/PollutedLocationDTOFileRepositoryTests.cs
```diff
@@ -270,5 +270,21 @@ public async Task GetByPropertyAsync_SeveralInstanceWithRequestedIdFound_FirstMa
...
+    {
+        var coordinates = new Coordinates
+        {
+            Longitude = 55.7329098,
+            Latitude = 24.3557527,
+        };
+       
+        var geocoderClass = new Geocoder();
+        var result = geocoderClass.ReverseGeocode(coordinates);
```
6ar8nas: use `await` with async method instead of result.Result
for example here it would be
```var result = await geocoderClass.ReverseGeocode(coordinates);```

File: Apsitvarkom.DataAccess/Geocoder.cs [hunk 8]
```diff
@@ -0,0 +1,35 @@
...
+using Newtonsoft.Json;
+
+
+namespace Apsitvarkom.DataAccess
+{
+    public class Geocoder : IGeocoder
+    {
+        private static readonly HttpClient client = new HttpClient();
+
+        public static string geocoderApiKey;
```
6ar8nas: I believe this is not the right way to do it.

Have a constructor, where you would take in the geocoderApiKey. I would also create the HttpClient in the constructor instead of its current location.

Apsitvarkom.DataAccess/Geocoder.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        public async Task<string> ReverseGeocode(Coordinates cords)
+        {
+            var apiResponse = await MakeReverseGeocodeApiCall(cords);
+            dynamic dynamicObject = JsonConvert.DeserializeObject<dynamic>(apiResponse)!;
+            string locationTitle = dynamicObject.results[0].formatted_address;
+            return locationTitle;
+        }
+
+        private static async Task<string> MakeReverseGeocodeApiCall(Coordinates cordinates)
```
6ar8nas:
```suggestion
        private static async Task<string> MakeReverseGeocodeApiCall(Coordinates coordinates)
```

Apsitvarkom.Api/Program.cs
```diff
@@ -6,6 +6,10 @@
 using FluentValidation;
 
 var builder = WebApplication.CreateBuilder(args);
+if (builder.Environment.IsLocal())
+{
+    builder.Configuration.AddUserSecrets<Program>();
+}
```
BenasB: This is no longer needed when you update your branch with changes from `staging`

Apsitvarkom.DataAccess/Geocoder.cs
```diff
@@ -0,0 +1,32 @@
+﻿using Apsitvarkom.Models;
+using JsonSerializer = System.Text.Json.JsonSerializer;
+
+namespace Apsitvarkom.DataAccess
+{
+    public class Geocoder : IGeocoder
+    {
+        private static HttpClient _client;
```
BenasB: No need to have `static` here (as a rule of thumb you should only use `static` when you *need* it, not because it *can* be used), also add `readonly` (you only assign it in the constructor)

Apsitvarkom.DataAccess/Geocoder.cs
```diff
@@ -0,0 +1,32 @@
...
+
+        private static string? _apiKey;
```
BenasB: No need for `static` (see previous comment why not) and nullable here (you *always* assign it in the constructor), also add `readonly`. So result should be:

`private readonly string _apiKey;`

Apsitvarkom.DataAccess/Geocoder.cs
```diff
@@ -0,0 +1,32 @@
...
+        public Geocoder(string apiKey)
+        {
+            _apiKey = apiKey;
+            _client = new HttpClient();
+        }
+        public async Task<string> ReverseGeocode(Coordinates coordinates)
+        {
+            var apiCall = _client.GetStreamAsync("https://maps.googleapis.com/maps/api/geocode/json?latlng=" + coordinates.Longitude + "," + coordinates.Latitude + "&language=lt&result_type=political&key=" + _apiKey);
```
BenasB: Use [string interpolation](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/interpolated) instead of adding strings with `+`

Example:
```csharp
int myVar = 5;
string a = "value is " + myVar + " and it is cool";
string b = $"value is {myVar} and it is cool"; // interpolated
```

BenasB: Move `https://maps.googleapis.com/maps/api/geocode/json` to a `const string GeocodeEndpoint` field in this class, since it will always be constant and might be reusing in the future.

Apsitvarkom.DataAccess/Geocoder.cs
```diff
@@ -0,0 +1,32 @@
...
+        }
+    }
+    public class ReverseGeocodingApiResult
+    {
+        public string? formatted_address { get; set; }
+    }
+    public class ReverseGeocodingApiResponse
+    {
+        public IEnumerable<ReverseGeocodingApiResult>? results { get; set; }
+    }
```
BenasB: These two classes should be `private` since they only matter in this `Geocoder` class

Apsitvarkom.DataAccess/Geocoder.cs
```diff
@@ -0,0 +1,32 @@
...
+        {
+            var apiCall = _client.GetStreamAsync("https://maps.googleapis.com/maps/api/geocode/json?latlng=" + coordinates.Longitude + "," + coordinates.Latitude + "&language=lt&result_type=political&key=" + _apiKey);
+            var deserializedApiResponse = await JsonSerializer.DeserializeAsync<ReverseGeocodingApiResponse>(await apiCall);
+            string formattedTitle = deserializedApiResponse.results.First().formatted_address;
+            return formattedTitle;
+        }
+    }
+    public class ReverseGeocodingApiResult
+    {
+        public string? formatted_address { get; set; }
```
BenasB: Please stick to the C# naming conventions, this property should be named `FormattedAddress`. `JsonConvert` probably has options/settings that takes into account naming and you can set it up, you will need to google for this

Apsitvarkom.DataAccess/Geocoder.cs
```diff
@@ -0,0 +1,32 @@
...
+    }
+    public class ReverseGeocodingApiResponse
+    {
+        public IEnumerable<ReverseGeocodingApiResult>? results { get; set; }
```
BenasB: Naming, same as previous comment

Apsitvarkom.DataAccess/IGeocoder.cs
```diff
@@ -0,0 +1,14 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Apsitvarkom.DataAccess
+{
+    public interface IGeocoder
+    {
+        Task<string> ReverseGeocode(Coordinates cords);
```
BenasB: `ReverseGeocodeAsync` instead of `ReverseGeocode` to indicate this is an asynchronous method

Apsitvarkom.Api/Controllers/GeocodingController.cs
```diff
@@ -0,0 +1,30 @@
...
+    {
+        private readonly IGeocoder _geocoder;
+
+        public GeocodingController(IGeocoder geocoder)
+        {
+            _geocoder = geocoder;
+        }
+        [HttpGet]
+        [ProducesResponseType(typeof(string), StatusCodes.Status200OK)]
+        public async Task<string> getGeocoder()
```
BenasB: PascalCase for methods, i. e. `GetGeocoder`. Even better would be `ReverseGeocode` since it describes what the action is used for

BenasB (changes requested):

Apsitvarkom.DataAccess/Geocoder.cs
XxDoMaSxX: "You can only restrict the access to the assembly containing the class using the internal modifier."

BenasB (approved):

BenasB:
Staging deployment will fail now since we haven't set up an api key for staging

PR #72, opened 2022-10-24
Get rid of Local environment, adjust pipelines

BenasB:
## What was done

- "Local" environment is now "Development" and "Development" is now "Staging"
- Adjusted pipelines
- Updated secrets in GitHub repo settings


BaVilius (approved):
Looks good

PR #76, opened 2022-10-25
Geocoder part 2

BenasB:
## What was done

- Added `Apsitvarkom.Configuration` project
- Clean up geocoder and rename the implementation to `GoogleGeocoder`
- Remove test controller
- Add 1 unit test (more should be done in a future task)
- Implement `ApiKeyProvider`


Apsitvarkom.DataAccess/IGeocoder.cs
```diff
@@ -1,14 +1,9 @@
 ﻿using Apsitvarkom.Models;
-using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
 
 namespace Apsitvarkom.DataAccess
```
6ar8nas:
```suggestion
namespace Apsitvarkom.DataAccess;
```

Apsitvarkom.DataAccess/GoogleGeocoder.cs
```diff
@@ -0,0 +1,42 @@
...
+
+    public GoogleGeocoder(HttpClient httpClient, IApiKeyProvider apiKeyProvider)
+    {
+        _apiKey = apiKeyProvider.Geocoding;
+        _httpClient = httpClient;
+    }
+
+    public async Task<string?> ReverseGeocodeAsync(Coordinates coordinates)
+    {
+        var jsonStream = await _httpClient.GetStreamAsync($"json?latlng={coordinates.Latitude},{coordinates.Longitude}&language=lt&result_type=political&key={_apiKey}");
```
6ar8nas, line 26: nice catch with the reversed coordinates

Apsitvarkom.DataAccess/GoogleGeocoder.cs
```diff
@@ -0,0 +1,42 @@
...
+        var response = await JsonSerializer.DeserializeAsync<ReverseGeocodingApiResponse>(jsonStream, SerializerOptions);
+
+        var title = response?.Results?.FirstOrDefault()?.FormattedAddress;
+        return title ?? string.Empty;
```
6ar8nas: Should we return an empty string in case we can't resolve the location? Maybe 'Unknown location' would be better for in-web usage, as the title would still exist, instead of it being simply blank?

Apsitvarkom.UnitTests/DataAccess/GoogleGeocoderTests.cs
```diff
@@ -0,0 +1,53 @@
+using System.Net;
+using System.Text;
```
6ar8nas: unused directive

Apsitvarkom.UnitTests/DataAccess/GoogleGeocoderTests.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        var handlerMock = new Mock<HttpMessageHandler>();
+        var response = new HttpResponseMessage
+        {
+            StatusCode = HttpStatusCode.OK,
+            Content = new StringContent(responseJsonString),
+        };
+
+        handlerMock
+            .Protected()
```
6ar8nas, line 27: seeing this protected for the first time. damn 🤯

6ar8nas (approved):
some small adjustments necessary. nice touches, saw some things for the first time.

feel free to resolve the comments on your own

Apsitvarkom.DataAccess/GoogleGeocoder.cs
BenasB: yeah it was returning UAE when in fact I wanted to get Vingio parkas

BenasB: There's also `language=lt` in the query. I am not sure how we want this handled. We could switch to `en`.

Or maybe we could come up with something more interesting and make the request in both languages (since we're planning to internationalise (add languages) to our front end)

Apsitvarkom.DataAccess/GoogleGeocoder.cs
BenasB: Yeah good comment, in fact we have the method declaration to return `string?` and currently we can't return `null` in any way.

Your suggestion makes sense at first, but here's why we shouldn't do that when you think about it: by adding a default value such as "Unknown location" we're assigning a value to a field that will probably get rendered by our API users (for example our front end). This poses a problem since we're putting it in a specific language. So if our front end has `lt` language support and receives this default value back, how should it render it? In english? Check if it's "Unknown location" and then replace it with a Lithuanian counterpart? What if the default value returned by the API changes?

So my suggestion is to return `null` if geocoding API can't find an address for those coordinates. This will let the API user decide how he wants to render a title that was not assigned. When we actually use this geocoder, it could be wise to log a warning if we can't find the address.

Apsitvarkom.UnitTests/DataAccess/GoogleGeocoderTests.cs
BenasB: yeah me too :)

BenasB: I haven't mocked an `HttpClient` before, so I think I followed this: https://dev.to/gautemeekolsen/how-to-test-httpclient-with-moq-in-c-2ldp

Apsitvarkom.DataAccess/GoogleGeocoder.cs
6ar8nas: internationalization sounds better imo.

BenasB: Let's keep that in mind and maybe we could create a task for that. Although we would need to figure out how to best do it and how it will be saved to the database

Apsitvarkom.DataAccess/GoogleGeocoder.cs
6ar8nas: Didn't consider the languages part at first, so I guess "Unknown location" <= string.Empty (could also technically make a check if string is empty) < null in this case

PR #81, opened 2022-10-31
Write additional unit tests for written GoogleGeocoder class

### LINKED ISSUE #77: More unit tests for GoogleGeocoder
## Description

Currently there is [one unit test for GoogleGeocoder](https://github.com/vu-vibedosa/apsitvarkom-be/blob/staging/Apsitvarkom.UnitTests/DataAccess/GoogleGeocoderTests.cs#L14). It would be good to test all of the possible paths.

## Steps to take

1.  If you are not familiar with Google's Geocoding API, read [this](https://developers.google.com/maps/documentation/geocoding/requests-reverse-geocoding)
2. Write tests and mock the different possible responses from the Geocoding API, such as
    1. No results found
    2. Only one result found
    3. Multiple results found
    4. ... 
3. Write tests for different scenarios you can come up with (e. g. multiple calls to `ReverseGeocode`, etc.)
4. As usual when writing tests, try to reuse mock objects between tests when you can, share common logic between tests

## Acceptance criteria

1. More unit tests written for `GoogleGeocoder`

6ar8nas:
## What was done

Written unit tests to cover different cases of interactions with google geocoder API.

### DIFF
--- a/Apsitvarkom.DataAccess/GoogleGeocoder.cs
+++ b/Apsitvarkom.DataAccess/GoogleGeocoder.cs
@@ -23,7 +23,7 @@ public class GoogleGeocoder : IGeocoder
 
     public async Task<string?> ReverseGeocodeAsync(Coordinates coordinates)
     {
-        var jsonStream = await _httpClient.GetStreamAsync($"json?latlng={coordinates.Latitude},{coordinates.Longitude}&language=lt&result_type=political&key={_apiKey}");
+        var jsonStream = await _httpClient.GetStreamAsync($"json?latlng={coordinates.Latitude},{coordinates.Longitude}&language=lt&result_type=point_of_interest&key={_apiKey}");
         var response = await JsonSerializer.DeserializeAsync<ReverseGeocodingApiResponse>(jsonStream, SerializerOptions);
 
         return response?.Results?.FirstOrDefault()?.FormattedAddress;
--- a/Apsitvarkom.DataAccess/GoogleGeocoder.cs
+++ b/Apsitvarkom.DataAccess/GoogleGeocoder.cs
@@ -23,7 +23,7 @@ public class GoogleGeocoder : IGeocoder
 
     public async Task<string?> ReverseGeocodeAsync(Coordinates coordinates)
     {
-        var jsonStream = await _httpClient.GetStreamAsync($"json?latlng={coordinates.Latitude},{coordinates.Longitude}&language=lt&result_type=point_of_interest&key={_apiKey}");
+        var jsonStream = await _httpClient.GetStreamAsync($"json?latlng={coordinates.Latitude},{coordinates.Longitude}&language=lt&key={_apiKey}");
         var response = await JsonSerializer.DeserializeAsync<ReverseGeocodingApiResponse>(jsonStream, SerializerOptions);
 
         return response?.Results?.FirstOrDefault()?.FormattedAddress;
--- a/Apsitvarkom.UnitTests/DataAccess/GoogleGeocoderTests.cs
+++ b/Apsitvarkom.UnitTests/DataAccess/GoogleGeocoderTests.cs
@@ -26,47 +26,46 @@ public class GoogleGeocoderTests
     }
 
     [Test]
-    public void OkStatusCodeReturned_SingleAddressRetrieved_TitleSuccessfullyReturned()
+    public Task OkStatusCodeReturned_SingleAddressRetrieved_TitleSuccessfullyReturned()
     {
         const string title = "Hello World";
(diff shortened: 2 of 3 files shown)


Apsitvarkom.DataAccess/GoogleGeocoder.cs
```diff
@@ -23,7 +23,7 @@ public GoogleGeocoder(HttpClient httpClient, IApiKeyProvider apiKeyProvider)
 
     public async Task<string?> ReverseGeocodeAsync(Coordinates coordinates)
     {
-        var jsonStream = await _httpClient.GetStreamAsync($"json?latlng={coordinates.Latitude},{coordinates.Longitude}&language=lt&result_type=political&key={_apiKey}");
+        var jsonStream = await _httpClient.GetStreamAsync($"json?latlng={coordinates.Latitude},{coordinates.Longitude}&language=lt&result_type=point_of_interest&key={_apiKey}");
```
BenasB: As we've discussed privately, I don't see the need to have `&result_type=point_of_interest` here, since if a location won't be associated with any point of interest, then it will return no results (although it could have give us just the street address). We will be picking out just the first result and that should be enough for us

Apsitvarkom.UnitTests/DataAccess/GoogleGeocoderTests.cs
```diff
@@ -1,48 +1,88 @@
...
 using Moq.Protected;
 
 namespace Apsitvarkom.UnitTests.DataAccess;
 
 public class GoogleGeocoderTests
 {
-    private static readonly IApiKeyProvider _apiKeyProvider = new Mock<IApiKeyProvider>().Object;
+    private Mock<IApiKeyProvider> _apiKeyProvider;
+    private Mock<HttpMessageHandler> _handlerMock;
+    private IGeocoder _geocoder;
```
BenasB: Missing `readonly` ?

Apsitvarkom.UnitTests/DataAccess/GoogleGeocoderTests.cs
```diff
@@ -1,48 +1,88 @@
...
+        const string responseJsonString = "{\"results\":[{\"formatted_address\":\"" + title + "\"}, {\"formatted_address\": \"second address\"}], \"status\": \"OK\"}";
 
-        var handlerMock = new Mock<HttpMessageHandler>();
+        return TestReverseGeocodeRequest(responseJsonString, title);
+    }
+
+    [Test]
+    public Task OkStatusCodeReturned_NullAddressRetrieved_NullReturned()
+    {
+        const string responseJsonString = "{\"results\":[{\"formatted_address\":null}], \"status\": \"OK\"}";
```
BenasB: I am not sure what adding `status` to the json result string gives us here. I've read [the docs](https://developers.google.com/maps/documentation/geocoding/requests-reverse-geocoding#reverse-status-codes) but that doesn't give us anything since we don't even take into account that field (in `GoogleGeocoder.cs` we only deserialize `results` -> `formatted_address`). You could write any other valid json and we would just ignore it just like we now ignore that `status` you put in the tests.

BenasB (changes requested):

Apsitvarkom.UnitTests/DataAccess/GoogleGeocoderTests.cs
6ar8nas: `readonly` properties can only be assigned in constructors and here I am using `SetUp` only.

Apsitvarkom.UnitTests/DataAccess/GoogleGeocoderTests.cs
6ar8nas: Practically for our use case it doesn't add nothing, good point. I have considered slightly altering what we parse from the response to include status too, but figured it would probably be redundant. I guess I just didn't remove it afterwards. Will delete

Apsitvarkom.UnitTests/DataAccess/GoogleGeocoderTests.cs
6ar8nas: similarly is done in [here](https://github.com/vu-vibedosa/apsitvarkom-be/blob/7f3f5a065c7409df19e1ab147e379d0d4f7ffabf/Apsitvarkom.UnitTests/Models/Mapping/PollutedLocationMappingTests.cs#L12) or [here](https://github.com/vu-vibedosa/apsitvarkom-be/blob/7f3f5a065c7409df19e1ab147e379d0d4f7ffabf/Apsitvarkom.UnitTests/DataAccess/PollutedLocationDTOFileRepositoryTests.cs#L19)

BenasB: ahh I forgot this was not the constructor lol

BenasB (approved):

Issue #82, opened 2022-11-05
Implement PollutedLocationDTODatabaseRepository


6ar8nas:
## Prerequisites

This task requires [infra issue no.13](https://github.com/vu-vibedosa/apsitvarkom-infra/issues/13) and #83 to be completed.

## Description

In [`PollutedLocationDTODatabaseRepository`](https://github.com/vu-vibedosa/apsitvarkom-be/blob/staging/Apsitvarkom.DataAccess/PollutedLocationDTODatabaseRepository.cs#L8) there are some not implemented methods. This task should implement them.

## Steps to take

1. Familiarize yourself with what was done in #83 and how we use EF in our project now (read ["Running locally"](https://github.com/vu-vibedosa/apsitvarkom-be/blob/staging/README.md#running-locally) to learn how to start the project now).
2. Implement 2 not implemented methods in `PollutedLocationDTODatabaseRepository`
3. Look into how EF `DbContext` can be mocked (we're using `Moq` to mock the dependencies, so google something among the lines of "C# moq mock entity framework dbcontext") and try to write some unit tests to cover `PollutedLocationDTODatabaseRepository` behavior.
4. Delete the previously used [JSON file](https://github.com/vu-vibedosa/apsitvarkom-be/blob/staging/Apsitvarkom.Api/PollutedLocationsMock.json) for storing of our PollutedLocationDTO instances. The content from it has been turned into [`DbInitializer`](https://github.com/vu-vibedosa/apsitvarkom-be/blob/staging/Apsitvarkom.DataAccess/DbInitializer.cs#L10)

## Acceptance criteria

1. Fully implemented `PollutedLocationDTODatabaseRepository` with unit testing.

BenasB:
vu-vibedosa/apsitvarkom-infra#13 was implemented, although it does not allow connections from the internet yet. I think we should start with a locally running database (the developer has the DB on their own machine for local development) - from there we can look into hooking up our API to the staging database.

Another point is that 1) we don't have a way to initialize schema/tables in the database (even the staging database is currently empty) and 2) we also don't have a good way of querying the data

Although 1 could be done with [FluentMigrator](https://fluentmigrator.github.io/) and 2 with [Dapper](https://github.com/DapperLib/Dapper), both of these steps should be possible with [Entity Framework](https://learn.microsoft.com/en-us/ef/) (which is a requirement). So my best bet is to start analyzing how entity framework works and trying to use that alongside a locally running database

[...]

PR #85, opened 2022-11-06
Setup Entity Framework for PollutedLocation

### LINKED ISSUE #83: Create DbContext for PollutedLocation
## Description

In preparation for #82, we need something that initializes the database schema and has a way to query the database (as mentioned in [this comment](https://github.com/vu-vibedosa/apsitvarkom-be/issues/82#issuecomment-1304633566)). Let's start using Entity Framework from the start because we'll need to use it eventually (it's a lab 3 requirement). 

## Steps to take

1. Either install PostgreSQL on your machine [directly](https://www.postgresql.org/download/) or [use Docker](https://hevodata.com/learn/docker-postgresql/)
2. Go through the [MS EF Razor pages tutorial](https://learn.microsoft.com/en-us/aspnet/core/data/ef-rp/intro?view=aspnetcore-6.0), it should have most of what we need (some parts we don't need because we don't use razor pages)
3. Implement a `DbContext` for `PollutedLocation` that is ready to be used by repository class (#82)
4. Make staging environment communicate with the staging database created in vu-vibedosa/apsitvarkom-infra#13

## Acceptance criteria

1. Implemented `DbContext` class for `PollutedLocation`
2. API starts in staging environment (is able to communicate with staging database)

BenasB:
## What was done

- Added `PollutedLocationsContext`
- Added `PollutedLocationDTODatabaseRepository`. This is now used instead of the file repository (left #82 to implement it fully and to also clean up unneeded things from file repository)
- Added `DbInitializer` that adds some fake data if the database doesn't have anything (for `development` and `staging` environments)
- Make `Notes` optional (as we've planned at the start)
- Switch `Coordinates` from `struct` to `class` (EF does not like structs)
- Updated readme with instructions on how to run the project now
- Allowed connection from the internet to our PostgreSQL VM. `staging` environment now connects to our `staging` database. Although this might not be fully sustainable, since we don't have a *static* external IP for the VM, so if it changes for some reason, we'll need to reset the connection string - so let's just wait and see how often that happens

Closes #83

### DIFF
--- a/Apsitvarkom.Api/Program.cs
+++ b/Apsitvarkom.Api/Program.cs
@@ -43,19 +43,16 @@ if (app.Environment.IsDevelopment() || app.Environment.IsStaging())
     app.UseSwagger();
     app.UseSwaggerUI();
 
-    using (var scope = app.Services.CreateScope())
-    {
-        var services = scope.ServiceProvider;
-        var pollutedLocationContext = services.GetRequiredService<PollutedLocationContext>();
+    using var scope = app.Services.CreateScope();
+    var services = scope.ServiceProvider;
+    var pollutedLocationContext = services.GetRequiredService<PollutedLocationContext>();
 
-        // TODO: switch to migrations
-        pollutedLocationContext.Database.EnsureCreated();
+    // TODO: switch to migrations
+    pollutedLocationContext.Database.EnsureCreated();
 
-        if (!pollutedLocationContext.PollutedLocations.Any())
-            DbInitializer.InitializePollutedLocations(pollutedLocationContext);
-    }
+    if (!pollutedLocationContext.PollutedLocations.Any())
+        DbInitializer.InitializePollutedLocations(pollutedLocationContext);
 }
-
 
 app.UseCors(FrontEndPolicy);
 app.UseAuthorization();
--- a/Apsitvarkom.Api/Program.cs
+++ b/Apsitvarkom.Api/Program.cs
@@ -2,7 +2,6 @@ using Apsitvarkom.Configuration;
 using Apsitvarkom.DataAccess;
 using Apsitvarkom.Models.DTO;
 using Apsitvarkom.Models.Mapping;
-using AutoMapper;
 using FluentValidation;
 using Microsoft.EntityFrameworkCore;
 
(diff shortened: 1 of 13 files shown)


BenasB:
`DbInitializer` messed up the coverage a lot haha 😄

Apsitvarkom.Api/Program.cs
```diff
@@ -44,8 +42,21 @@
...
+
+        // TODO: switch to migrations
+        pollutedLocationContext.Database.EnsureCreated();
+
+        if (!pollutedLocationContext.PollutedLocations.Any())
+            DbInitializer.InitializePollutedLocations(pollutedLocationContext);
+    }
 }
 
+
```
6ar8nas: seems like an accidental \n

Apsitvarkom.Api/Program.cs
```diff
@@ -44,8 +42,21 @@
 {
     app.UseSwagger();
     app.UseSwaggerUI();
+
+    using (var scope = app.Services.CreateScope())
```
6ar8nas: starting C# 8.0 you can also use using [without braces](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-8.0/using)     
`using var scope = app.Services.CreateScope();`

README.md
```diff
@@ -2,18 +2,23 @@
 
 Apsitvarkom back end dotnet web API
```
6ar8nas, line 3: Thanks for changing this one, especially including all the previously implemented and not documented changes to the setup process as well.

6ar8nas (approved):
Could setup the project with updated readme instructions.

Don't forget to update the #82 when you'd be free.

Apsitvarkom.Api/Program.cs
BenasB: Yeah I know, I just wanted to keep the scope of that using block clear. Does this generate a warning for you?

BenasB: I will just change it to the new syntax

6ar8nas: > Yeah I know, I just wanted to keep the scope of that using block clear. Does this generate a warning for you?

No, it didn't. Just figured as it was { ... using() {} } we could have simplified, as the block ended right away anyways.

PR #87, opened 2022-11-12
Implement PollutedLocationDTODatabaseRepository

### LINKED ISSUE #82: Implement PollutedLocationDTODatabaseRepository

6ar8nas:
## What was done

- Implemented missing interface methods
- Introduced an interface for PollutedLocationContext, so that I could mock it.
- Introduced one NuGet package to allow simpler unit testing of the db repo.
- Covered the implementation with tests
- Split DbInitializer to two methods, so that I can reuse the fake array creation and 'cover' that part of code.

Closes #82


Apsitvarkom.DataAccess/PollutedLocationContext.cs
```diff
@@ -3,7 +3,12 @@
 
 namespace Apsitvarkom.DataAccess;
 
-public class PollutedLocationContext : DbContext
+public interface IPollutedLocationContext
+{
+    DbSet<PollutedLocation> PollutedLocations { get; set; }
```
BenasB: Can we not have `set;` here? I don't think we're supposed to be able to set the `DbSet` from outside. Or `private set;`

Apsitvarkom.DataAccess/PollutedLocationDTODatabaseRepository.cs
```diff
@@ -1,29 +1,47 @@
...
+        return from pollutedLocation in await _context.PollutedLocations.ToListAsync()
+            orderby inRelationTo.DistanceTo(pollutedLocation)
+            select _mapper.Map<PollutedLocation, PollutedLocationDTO>(pollutedLocation);
+    }
 
-    public Task<PollutedLocationDTO?> GetByPropertyAsync(Func<PollutedLocationDTO, bool> propertyCondition) => throw new NotImplementedException();
+    /// <inheritdoc />
+    public async Task<PollutedLocationDTO?> GetByPropertyAsync(Func<PollutedLocationDTO, bool> propertyCondition)
+    {
+        var allLocations = await GetAllAsync();
```
BenasB: You should not depend on `GetAllAsync` here. AFAIK, EF's `DbSet` works by executing SQL queries when you need the result list (e.g. calling `ToListAsync`) or in other words when you move away from `IQueryable<T>` to a (for example) `List<T>`. So as a rule of thumb, you should keep `IQueryable<T>` as far as possible, because that will mean a more concrete SQL query.

So right now in this method this is happening: you are firstly executing SQL query that `SELECT`s all (by calling method `GetAllAsync`) and then doing `FirstOrDefault` on an `IEnumerable` (not `IQueryable`):

image

This is not optimal. Of course it works okay when we have a small number of rows in the database, but imagine we have a million of rows and this method would firstly query all of those rows and only then search for the first one. The optimal way would be to construct the SQL query to look for that one element immediately (without select all of the rows beforehand). You do that by chaining extension methods on `IQueryable`

BenasB: Another problem in our design here is that the `propertyCondition` is of type `Func<PollutedLocationDTO, bool>` but our `DbSet` is of type `PollutedLocation`, so you will also need to map between those types in the `FirstOrDefaultAsync` method.

We should think about this. I'm even more tempted to switch all of these repositories from DTO to business types and the controllers would just map the types themselves.

Apsitvarkom.DataAccess/PollutedLocationDTODatabaseRepository.cs
```diff
@@ -1,29 +1,47 @@
...
     }
 
-    public Task<IEnumerable<PollutedLocationDTO>> GetAllAsync(Location inRelationTo) => throw new NotImplementedException();
+    /// <inheritdoc />
+    public async Task<IEnumerable<PollutedLocationDTO>> GetAllAsync(Location inRelationTo)
+    {
+        return from pollutedLocation in await _context.PollutedLocations.ToListAsync()
+            orderby inRelationTo.DistanceTo(pollutedLocation)
+            select _mapper.Map<PollutedLocation, PollutedLocationDTO>(pollutedLocation);
+    }
```
BenasB, line 41: Same in this method as the previous comment, utilize `IQueryable` because right now you're just executing SQL query by calling `ToListAsync`. Preferably you would construct the query and then call `ToListAsync` at the end

Apsitvarkom.UnitTests/DataAccess/PollutedLocationDTODatabaseRepositoryTests.cs
```diff
@@ -0,0 +1,194 @@
...
+    [Test]
+    public void PollutedLocationDTODatabaseRepositoryConstructor_HappyPath() =>
+        Assert.DoesNotThrow(() => new PollutedLocationDTODatabaseRepository(m_mockContext.Object, m_mapper));
+    #endregion
+
+    #region GetAllAsync tests
+    [Test]
+    public async Task GetAllAsync_DbContextHasNoInstances_EmptyListReturned()
+    {
+        var dbInstances = new List<PollutedLocation>();
```
BenasB: Nit: I'd prefer a more descriptive name such as `dbRows` since I am not sure what `dbInstances` refer to here without seeing how it's used

Apsitvarkom.UnitTests/DataAccess/PollutedLocationDTODatabaseRepositoryTests.cs
```diff
@@ -0,0 +1,194 @@
...
+    public void SetUp()
+    {
+        var config = new MapperConfiguration(cfg =>
+        {
+            cfg.AddProfile<PollutedLocationProfile>();
+        });
+        config.AssertConfigurationIsValid();
+        m_mapper = config.CreateMapper(); 
+        m_mockContext = new Mock<IPollutedLocationContext>();
+    }
```
BenasB: Would be possible to extract some things into `[OneTimeSetUp]`, right? `m_mapper` does not change/mutate between tests

Apsitvarkom.UnitTests/DataAccess/PollutedLocationDTODatabaseRepositoryTests.cs
```diff
@@ -0,0 +1,194 @@
...
+        var notes = "notes";
+        var spotted = new DateTime(2022, 11, 12, 19, 23, 30);
+        var coordinates = new Coordinates { Latitude = 47.12, Longitude = -41.1251 };
+        var radius = 35;
+        var progress = 67;
+        var severity = LocationSeverityLevel.Moderate;
+        var dbInstances = new List<PollutedLocation>
+        {
+            new() { Id = id, Coordinates = coordinates, Radius = radius, Severity = severity, Spotted = spotted, Notes = notes, Progress = progress }
+        };
```
BenasB: Just wondering why is everything a separate variable? 😄 and then in line 61 you just have a lot of repeated variable names. I would just create the object inline (inside the `new()` in `new List<PollutedLocation>{}`)

BenasB: Or just create a single `new PollutedLocation` and add it to the `dbInstances` list (I just saw that you're asserting it later)

Apsitvarkom.DataAccess/DbInitializer.cs
```diff
@@ -9,7 +9,15 @@ public static class DbInitializer
...
     public static void InitializePollutedLocations(PollutedLocationContext context)
     {
-        var fakePollutedLocations = new PollutedLocation[]
+        var fakePollutedLocations = CreateFakePollutedLocationArray();
+
+        context.PollutedLocations.AddRange(fakePollutedLocations);
+        context.SaveChanges();
+    }
+
+    public static PollutedLocation[] CreateFakePollutedLocationArray()
```
BenasB: This is just a getter, right? No need for a whole method:

`public static PollutedLocation[] FakePollutedLocationArray => new PollutedLocation[] {...}`

(no need for `Create` in the name as well)

BenasB (changes requested):
Some improvements needed, but a welcome change. Good stuff with the EF mocking, I don't have experience with it yet, but it seems like a fitting tool.

Apsitvarkom.DataAccess/DbInitializer.cs
6ar8nas: changed it to be a lazy instead to fulfill the requirement, so sticked with the now private static method instead of it simply being an array

Apsitvarkom.DataAccess/PollutedLocationContext.cs
6ar8nas: we can't have no setter as the `PollutedLoccations` property then is equal to null and hence Program.cs line `pollutedLocationContext.PollutedLocations.Any()` throws.
we can however use the private setter, so I did just that.

Apsitvarkom.DataAccess/PollutedLocationDTODatabaseRepository.cs
6ar8nas: as we are using a DistanceTo method here, a simple SQL query to `order by` cannot resolve our method and we first have to get all the instances to list or to enumerable.
More information [here](https://learn.microsoft.com/en-us/ef/core/querying/client-eval).

image

Apsitvarkom.Api/Program.cs
```diff
@@ -21,6 +21,7 @@
     options.AddPolicy(FrontEndPolicy, policy => policy.WithOrigins(builder.Configuration.GetValue<string>("FrontEndOrigin")));
 });
 
+builder.Services.AddScoped<IPollutedLocationContext>(provider => provider.GetService<PollutedLocationContext>()!);
```
BenasB, line 24: Why not `GetRequiredService` instead of `GetService` ?

Apsitvarkom.DataAccess/PollutedLocationContext.cs
BenasB: yeah the whole point is that the interface does not have the `set`, so all good now

Apsitvarkom.DataAccess/PollutedLocationDTODatabaseRepository.cs
```diff
@@ -1,29 +1,51 @@
...
 
-    public Task<PollutedLocationDTO?> GetByPropertyAsync(Func<PollutedLocationDTO, bool> propertyCondition) => throw new NotImplementedException();
+    /// <inheritdoc />
+    public async Task<PollutedLocationDTO?> GetByPropertyAsync(Expression<Func<PollutedLocationDTO, bool>> propertyCondition)
+    {
+        return await _context.PollutedLocations
+            .ProjectTo<PollutedLocationDTO>(_mapper.ConfigurationProvider)
```
BenasB, line 47: Cool, didn't know about `ProjectTo`

Apsitvarkom.Api/Program.cs
BenasB: Why not `GetRequiredService` instead of `GetService` ?

Apsitvarkom.DataAccess/PollutedLocationDTODatabaseRepository.cs
BenasB: yeah that makes sense since we have C# code in that order by. I guess C# LINQ and SQL worlds clash here and there isn't a nice way (at least how we got things set up right now). We could investigate this in the future, but good enough for now I guess

BenasB (changes requested):

BenasB:
By the, the `Expression<T>` was added because we can make `async` stuff when filtering/querying, right?

Apsitvarkom.Api/Program.cs
6ar8nas: Good point. Didn't consider this at all. For any others that would read this and wouldn't be aware of the difference, more information [here](https://andrewlock.net/the-difference-between-getservice-and-getrquiredservice-in-asp-net-core/).

6ar8nas:
> By the, the Expression<T> was added because we can make async stuff when filtering/querying, right?

Yup, in this case particularly querying, as x => func(x) (which returns an expression) also throws the same error as the orderby statement with additional method/logic.

Apsitvarkom.Api/Program.cs
BenasB: I think you also don't need the `!` at the end then

BenasB (approved):

PR #92, opened 2022-11-14
Get validators from DI instead of newing

BenasB:
## What was done

- Get validators in other validators from DI instead of `new`ing
- Update `Spotted` mapping to get this format `"2022-06-08T01:12:23.0000000Z"`


6ar8nas (approved):
Nice catch. Didn't even notice the datetime string being like that. :D

PR #94, opened 2022-11-14
Added InformationAmountOfBytesPrinted event to FileLogger

BenasB:
## What was done

This fulfills

> 5. Events and their usage: standard and custom;

- Added `InformationAmountOfBytesPrinted` event that logs when a configurable amount of bytes is written to file


6ar8nas (approved):
I believe this is a temporary change, so looks good for the cause. We'll consider deleting this after. :D

PR #97, opened 2022-11-14
Extend integration testing with in-memory database repository testing

### LINKED ISSUE #96: Extend integration testing with in-memory tests
## Description

There is a requirement to have some unit and integration tests.
We've got the unit tests part covered; however, we do not have many integration tests present. Let's write some in-memory database integration tests to test how our repository works against a real database.

## Steps to take

1. Read more about in-memory testing [here](https://learn.microsoft.com/en-us/ef/core/testing/) and [here](https://stackoverflow.com/questions/54219742/mocking-ef-core-dbcontext-and-dbset/54220067#54220067).
2. Write tests for all of the existent `PollutedLocationDTODatabaseRepository` class methods.

## Acceptance criteria

At least one test for each method of the repository class is written and passing.

6ar8nas:
## What was done

- Slightly adjusted mapper to return correct result without regarding to timezones.
- Wrote an integration test for each of the Database Repository method.

closes #96

### DIFF
--- a/Apsitvarkom.IntegrationTests/Apsitvarkom.IntegrationTests.csproj
+++ b/Apsitvarkom.IntegrationTests/Apsitvarkom.IntegrationTests.csproj
@@ -9,6 +9,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.EntityFrameworkCore.InMemory" Version="6.0.11" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.1.0" />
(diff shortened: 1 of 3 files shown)


BenasB (approved):
Cool, thanks. Although it's worth reading [this](https://learn.microsoft.com/en-us/ef/core/testing/choosing-a-testing-strategy#in-memory-as-a-database-fake). In the future, we should switch to a real database using [testcontainers-dotnet](https://github.com/testcontainers/testcontainers-dotnet)

PR #100, opened 2022-11-17
Restructure DTOs into Responses and Requests

### LINKED ISSUE #99: Restructure DTOs
## Description

Restructure current DTOs to be endpoint/action specific following the request/response. [Something](https://stackoverflow.com/a/44350309/13135665) to think about. Also, from now on, those request/response classes should have non nullable types where needed (basically represent a validated object), unlike previously we had only nullables. We are keeping business object the same, it's fine.

## Steps to take

1. Restructure DTOs, create missing ones. This includes making properties non nullable
2. Switch back to composition instead of inheritance in `PollutedLocation` - `Location`
3. Update places where DTOs were used, repositories
4. Validators, mappers adjusted
5. Swagger integration with FluentValidation

## Acceptance criteria

1. No more `DTO`s, and request/response classes instead of them
2. Swagger now shows required fields based on FluentValidation validators instead of the default asp.net core validators

BenasB:
## What was done

- Restructured `DTO`s
    - The naming is replaced to `Response`s and `Request`s. So the string `DTO` should not appear anywhere in the codebase.
    - **Everything in `Apsitvarkom.Models/Public` is a public facing model** (they tell the users of our API the schema of what to expect from our endpoints) so anything outside of `Apsitvarkom.Models/Public` should not be exposed by a controller.
    - Please note that requests/responses should now have non primitive types if needed (e.g. `DateTime`, `Guid`) unlike previously where we had only primitives, this is due to swagger integration.
    - Adjusted validation and mapping. _Both requests and responses_ have validators (in theory we **don't call validators on the responses**, but they still serve a purpose to inform about the schema in swagger).
    - Since this is a big PR, you'll probably want to start from there.
- Switched from inheritance to composition for `Location` (like we used to have in the beginning before changing it due to a requirement) in our models
- `Repository` now acts on business objects. 
    - Validation, mapping should be performed by the controller. 
    - Removed `FileRepository` due to inconsistencies with out schema and not needing it in general.
- Adjusted `PollutedLocationController` actions
    -  `All/OrderedByDistance` now takes in a `CoordinatesCreateRequest` instead of two separate primitive `double`s
    -  `try/catch`es
- `DistanceTo` is now a method in `Coordinates` instead of an extension method
- Adjusted a lot of tests, removed some unneeded 
- Added FluentValidation integration with Swagger. This allows us to automatically generate schema information for swagger:
image

Closes #99


Apsitvarkom.Api/Apsitvarkom.Api.csproj
```diff
@@ -10,6 +10,7 @@
   <ItemGroup>
     <PackageReference Include="AutoMapper.Extensions.Microsoft.DependencyInjection" Version="11.0.0" />
     <PackageReference Include="FluentValidation.DependencyInjectionExtensions" Version="11.2.2" />
+    <PackageReference Include="MicroElements.Swashbuckle.FluentValidation" Version="6.0.0-beta.1" />
```
6ar8nas, line 13: I don't think this might cause any issues at our application level, but do we need to use 6.x for any reason in particular, or would a stable version (5.7 I believe) do just fine as well?

Apsitvarkom.Api/Program.cs
```diff
@@ -1,23 +1,29 @@
...
+        options.JsonSerializerOptions.Converters.Add(new JsonStringEnumConverter())
+    );
 builder.Services.AddEndpointsApiExplorer();
 builder.Services.AddSwaggerGen();
 
 builder.Services
     .AddAutoMapper(typeof(PollutedLocationProfile))
-    .AddValidatorsFromAssemblyContaining<PollutedLocationDTOValidator>();
+    .AddValidatorsFromAssemblyContaining<CoordinatesGetRequestValidator>()
+    .AddFluentValidationRulesToSwagger();
```
6ar8nas, line 37: that's a neat touch, nice.

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
```diff
@@ -1,6 +1,9 @@
 ﻿using Apsitvarkom.DataAccess;
 using Apsitvarkom.Models;
-using Apsitvarkom.Models.DTO;
+using Apsitvarkom.Models.Public;
+using AutoMapper;
+using FluentValidation;
+using FluentValidation.Results;
```
6ar8nas: unused

Apsitvarkom.DataAccess/PollutedLocationDatabaseRepository.cs
```diff
@@ -0,0 +1,40 @@
...
+    {
+        return from pollutedLocation in await _context.PollutedLocations.ToListAsync()
+            orderby pollutedLocation.Location.Coordinates.DistanceTo(inRelationTo)
+            select pollutedLocation;
+    }
+
+    /// <inheritdoc />
+    public Task<PollutedLocation?> GetByPropertyAsync(Expression<Func<PollutedLocation, bool>> propertyCondition)
+    {
+        return _context.PollutedLocations.FirstOrDefaultAsync(propertyCondition);
```
6ar8nas, line 38: true, could have used it instead of `.Where(cond).FirstOrDefaultAsync()` right off the bat... :D nice

Apsitvarkom.Models/Apsitvarkom.Models.csproj
```diff
@@ -7,7 +7,6 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <Folder Include="DTO\" />
```
6ar8nas, line 10: I guess it wasn't you who changed this file, but shouldn't we have the new `Public` folder here?

Apsitvarkom.Models/Public/PollutedLocationGetResponse.cs
```diff
@@ -0,0 +1,32 @@
...
+
+    public SeverityLevel Severity { get; set; }
+
+    public DateTime Spotted { get; init; }
+
+    public int Progress { get; set; }
+
+    public string? Notes { get; set; }
+
+    public LocationGetResponse Location { get; set; } = new();
```
6ar8nas: I would maybe put this property higher above - as we've previously discussed this determines the position in swagger too and I believe Location is one of the most important things in this response.

6ar8nas: As this is the initial implementation of using the new public classes instead of DTO objects, other similarly created responses might repeat this order too.

Apsitvarkom.UnitTests/Api/Controllers/PollutedLocationControllerTests.cs
```diff
@@ -2,7 +2,10 @@
 using Apsitvarkom.Api.Controllers;
 using Apsitvarkom.DataAccess;
 using Apsitvarkom.Models;
-using Apsitvarkom.Models.DTO;
+using Apsitvarkom.Models.Mapping;
+using Apsitvarkom.Models.Public;
+using AutoMapper;
+using FluentValidation.Results;
```
6ar8nas: not used

Apsitvarkom.UnitTests/Api/Controllers/PollutedLocationControllerTests.cs
```diff
@@ -49,71 +59,118 @@ public class PollutedLocationControllerTests
...
-        const double longitude = -65.4321;
-        m_repository.Setup(self => self.GetAllAsync(It.Is<Location>(x => 
-            Math.Abs(x.Coordinates.Latitude - latitude) < 0.0001 && Math.Abs(x.Coordinates.Longitude - longitude) < 0.0001))).ReturnsAsync(PollutedLocationDTOs);
+        var coordinatesGetRequest = new CoordinatesGetRequest
+        {
+            Latitude = 12.3456,
+            Longitude = -65.4321
+        };
+
+        m_repository.Setup(self => self.GetAllAsync(It.Is<Coordinates>(
```
6ar8nas: FYI in no way is that a mistake and that's a good point, that I shouldn't have put a one liner this long, but I would have instead formatted it like this:
```
m_repository.Setup(self => self.GetAllAsync(It.Is<Coordinates>(x => 
        Math.Abs((double)(x.Latitude - coordinatesGetRequest.Latitude)) < 0.0001 &&
        Math.Abs((double)(x.Longitude - coordinatesGetRequest.Longitude)) < 0.0001
    )))
    .ReturnsAsync(PollutedLocations);
```

this way seeing the trend here is simpler

Apsitvarkom.UnitTests/Api/Controllers/PollutedLocationControllerTests.cs
```diff
@@ -49,71 +59,118 @@ public class PollutedLocationControllerTests
...
-    public async Task GetAll_RepositoryReturnsPollutedLocationDTOs_OKActionResultReturned()
+    public async Task GetAll_RepositoryReturnsPollutedLocations_OKActionResultReturned()
     {
-        m_repository.Setup(self => self.GetAllAsync()).ReturnsAsync(PollutedLocationDTOs);
+        m_repository.Setup(self => self.GetAllAsync())
+            .ReturnsAsync(PollutedLocations);
 
         var actionResult = await m_controller.GetAll();
 
+        Assert.That(actionResult.Result, Is.TypeOf <OkObjectResult>());
```
6ar8nas:
```suggestion
        Assert.That(actionResult.Result, Is.TypeOf<OkObjectResult>());
```
you have written it without the space elsewhere:D

Apsitvarkom.UnitTests/Api/Controllers/PollutedLocationControllerTests.cs
```diff
@@ -49,71 +59,118 @@ public class PollutedLocationControllerTests
...
 
         var actionResult = await m_controller.GetById(instanceId);
 
+        Assert.That(actionResult.Result, Is.TypeOf<NotFoundObjectResult>());
         var result = actionResult.Result as NotFoundObjectResult;
 
         Assert.That(result, Is.Not.Null);
-        Assert.That(result!.StatusCode, Is.EqualTo(StatusCodes.Status404NotFound));
+        Assert.That(result.StatusCode, Is.EqualTo(StatusCodes.Status404NotFound));
+        Assert.That(result.Value, Is.Not.Null.And.Not.Empty);
```
6ar8nas, line 239: first time i'm seeing this :D nice one

Apsitvarkom.UnitTests/DataAccess/PollutedLocationDatabaseRepositoryTests.cs
```diff
@@ -182,18 +167,18 @@ public async Task GetByPropertyAsync_SeveralInstancesWithPropertyFound_FirstMatc
         var dbRows = new List<PollutedLocation>
         {
             new() { Id = Guid.NewGuid() },
-            new() { Id = id, Severity = LocationSeverityLevel.Low },
+            new() { Id = id, Severity = PollutedLocation.SeverityLevel.Low },
             new() { Id = Guid.NewGuid() },
-            new() { Id = Guid.NewGuid(), Severity = LocationSeverityLevel.Low}
+            new() { Id = Guid.NewGuid(), Severity = PollutedLocation.SeverityLevel.Low}
```
6ar8nas:
```suggestion
            new() { Id = Guid.NewGuid(), Severity = PollutedLocation.SeverityLevel.Low }
```
this one's on my previously done changes, but as I've noticed it now, might as well change it here I guess

Apsitvarkom.UnitTests/DataAccess/PollutedLocationDatabaseRepositoryTests.cs
```diff
@@ -58,17 +41,19 @@ public async Task GetAllAsync_DbContextHasOneInstance_SingleInstanceReturnedWith
         var pollutedLocationInstance = new PollutedLocation 
         { 
             Id = Guid.NewGuid(),
-            Coordinates = new Coordinates { Latitude = 47.12, Longitude = -41.1251 }, 
+            Location = {
```
6ar8nas: this seems to be the only place where we have the curly bracket in the same line as Location.

BenasB:
Please also check the FE PR for schema changes

6ar8nas (approved):
That's nice. Didn't know how difficult the implementation of this would have been when discussing in real life, but this does really seem to simplify things up. 

Wrote some comments, some of them don't even require changes. Feel free to resolve them by yourself.

Might make a quick PR resolving some things I've noticed while looking at this one here. Will close the comments related to what I'll have covered there.

Apsitvarkom.Api/Apsitvarkom.Api.csproj
BenasB: yes, this needs to be a prerelease, see https://github.com/micro-elements/MicroElements.Swashbuckle.FluentValidation/issues/111

Apsitvarkom.DataAccess/PollutedLocationDatabaseRepository.cs
BenasB: Yeah I'm not sure how we didn't notice it before

Apsitvarkom.Models/Public/PollutedLocationGetResponse.cs
BenasB: Sure

6ar8nas (approved):
looks good

Issue #90, opened 2022-11-14
Extend IDTORepository and PollutedLocationController with Delete


6ar8nas:
## Description

A simple instance delete CRUD action is necessary so that we could offer the ability for the user to remove an instance from the polluted location list. Let's join this task altogether with the next and implement this method straight into the controller as well. Use contents of #101 as a reference.

## Steps to take

1. Extend `IDTORepository` with `Delete` method taking the PollutedLocationDeleteRequest (it would consist only of the Id) the instance to-be-deleted as a parameter. Create their validators and mappers.
2. Implement the method in our `PollutedLocationDTODatabaseRepository` class.
3. Initiate a discussion for if deleted instances should be saved in some archive. Consider what should be done if the instance by the given id does not exist in the database. 
4. Implement the newly created method in the PollutedLocationController class. Take the needed data from the URI.
5. Write unit tests to cover the implemented functionality in the controller, validator, mapper and repository.

## Acceptance criteria

You can delete an instance using a delete request in swagger, with the endpoint at PollutedLocation/Delete. The implementation is covered with unit tests.

PR #101, opened 2022-11-17
Create PollutedLocation action

### LINKED ISSUE #88: Extend IDTORepository and PollutedLocationController with Create
## Description

A simple instance insertion CRUD action is necessary so that we could use it to `Report New` polluted locations in the front-end.
Let's join this task altogether with the next and implement this method straight into the controller as well.

## Steps to take

1. Extend `IDTORepository` with `Create` method taking the DTO object to be inserted as a parameter.
2. Implement the method in our `PollutedLocationDTODatabaseRepository` class (you can throw `NotImplementedException` in `PollutedLocationDTOFileRepository`).
3. Consider which properties the DTO object should be mandatory for the creation to be successful.
4. Use Geocoder to store the instance's title to the database.
5. Implement the newly created method in the PollutedLocationController class. Take the instance from post body.
6. Write unit tests to cover the implemented functionality in both the controller and repository.

## Acceptance criteria

You can create an instance using a post request in swagger, with the endpoint being PollutedLocation/Create. The implementation is covered with unit tests.

BenasB:
## What was done

Implemented a new `Create` action in `PollutedLocationController` based on the new request/response setup (this PR merges not to `staging`).

- Created necessary `CreateRequest` models, namely `PollutedLocationCreateRequest`. Created their validators and mappings.
- Renamed `CoordinatesGetRequest` to `CoordinatesCreateRequest` since it makes more sense, I should've given it the same name back in #100
- Added a new action to `PollutedLocationController`
- Implemented `InsertAsync` in the repository. Also needed to update the interface a bit.
- Written additional tests

Closes #88


Apsitvarkom.Api/Controllers/PollutedLocationController.cs
```diff
@@ -43,7 +52,7 @@ public async Task<ActionResult<IEnumerable<PollutedLocationGetResponse>>> GetAll
     [ProducesResponseType(StatusCodes.Status400BadRequest, Type = typeof(List<string>))]
     [ProducesResponseType(StatusCodes.Status500InternalServerError)]
     [HttpGet("All/OrderedByDistance")]
-    public async Task<ActionResult<IEnumerable<PollutedLocationGetResponse>>> GetAll([FromQuery] CoordinatesGetRequest coordinates)
+    public async Task<ActionResult<IEnumerable<PollutedLocationGetResponse>>> GetAll([FromQuery] CoordinatesCreateRequest coordinates)
```
6ar8nas: this seems odd, as we use it for get action. Maybe let's simply call it `CoordinatesRequest`, as I don't imagine such a scenario where a different coordinate request would exist.

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
```diff
@@ -71,7 +84,41 @@ public async Task<ActionResult<PollutedLocationGetResponse>> GetById(string id)
...
+        {
+            return StatusCode(StatusCodes.Status500InternalServerError);
+        }
+    }
+
+    [ProducesResponseType(StatusCodes.Status201Created)]
+    [ProducesResponseType(StatusCodes.Status400BadRequest, Type = typeof(List<string>))]
+    [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+    [HttpPost("Create")]
+    public async Task<ActionResult<PollutedLocationGetResponse>> Create(PollutedLocationCreateRequest pollutedLocationCreateRequest)
```
6ar8nas: same here with `PollutedLocationGetResponse` - aren't we always going to return the whole PollutedLocation instance? in that case, perhaps we could call it `PollutedLocationResponse`

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
```diff
@@ -71,7 +84,41 @@ public async Task<ActionResult<PollutedLocationGetResponse>> GetById(string id)
...
+
+        var mappedPollutedLocation = _mapper.Map<PollutedLocation>(pollutedLocationCreateRequest);
+        if (mappedPollutedLocation is null) return StatusCode(StatusCodes.Status500InternalServerError);
+
+        try
+        {
+            var result = await _repository.InsertAsync(mappedPollutedLocation);
+
+            var response = _mapper.Map<PollutedLocationGetResponse>(result);
+            // If failed to map the response, still return 201 because the location was inserted into the database
```
6ar8nas: If we can't map it to response, there must have been something wrong with the data I believe. Shouldn't we delete the inserted location in this case (would require #90) instead of potentially having it junked with wrong data?

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
```diff
@@ -31,7 +36,11 @@ public async Task<ActionResult<IEnumerable<PollutedLocationGetResponse>>> GetAll
         try
         {
             var locations = await _repository.GetAllAsync();
-            return Ok(locations);
+
+            var mappedLocations = _mapper.Map<IEnumerable<PollutedLocationGetResponse>>(locations);
```
6ar8nas: how does this differ from the unchanged version? are there cases where it matters? it seems like the AutoMapper handles this for us, if I'm not mistaken.

6ar8nas: (commenting before having looked at the tests)

Apsitvarkom.Api/Program.cs
```diff
@@ -18,11 +19,25 @@
         options.JsonSerializerOptions.Converters.Add(new JsonStringEnumConverter())
     );
 builder.Services.AddEndpointsApiExplorer();
-builder.Services.AddSwaggerGen();
+builder.Services.AddSwaggerGen(c =>
+{
```
6ar8nas, line 23: :D niice

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
6ar8nas: this would also make it so that the same status code can return two completely different responses content wise.

Apsitvarkom.DataAccess/IRepository.cs
```diff
@@ -15,4 +15,9 @@ public interface IRepository<T> where T: class
     /// <param name="propertyCondition">Condition for the requested record acquisition.</param>
     /// <returns>Null if the instance was not found. Otherwise the acquired <see cref="T" /> instance.</returns>
     public Task<T?> GetByPropertyAsync(Expression<Func<T, bool>> propertyCondition);
+
+    /// <summary>Inserts a single <see cref="T" /> record to the data source.</summary>
+    /// <param name="modelToInsert">Model to be inserted into the repository.</param>
+    /// <returns>The inserted <see cref="T" /> model.</returns>
+    public Task<T> InsertAsync(T modelToInsert);
```
6ar8nas: Why do we need this method to return the same object we pass to it?

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
```diff
@@ -71,7 +84,41 @@ public async Task<ActionResult<PollutedLocationGetResponse>> GetById(string id)
...
+
+        try
+        {
+            var result = await _repository.InsertAsync(mappedPollutedLocation);
+
+            var response = _mapper.Map<PollutedLocationGetResponse>(result);
+            // If failed to map the response, still return 201 because the location was inserted into the database
+            if (response is null) return StatusCode(StatusCodes.Status201Created);
+
+            return CreatedAtAction(nameof(GetById), new { response.Id }, response);
```
6ar8nas: This also doesn't seem necessary. We have `PollutedLocationGetResponse` at our hands. Why do we need to use a query to get it again?

Apsitvarkom.IntegrationTests/DataAccess/DatabaseTests.cs
```diff
@@ -73,4 +73,17 @@ public async Task GetByPropertyTest()
             Assert.That(response.Notes, Is.EqualTo(dbRow.Notes));
         });
     }
+
+    [Test]
+    public async Task InsertAsync_SamePrimaryKeys_Throws()
```
6ar8nas, line 78: We could also test for successful addition.

6ar8nas (changes requested):
Didn't look into tests as of submitting this piece of feedback, as changes might be needed, which would result in also altering the unit tests.

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
BenasB: This is `CoordinatesCreateRequest` because it seemed logical to reuse the same object that is in the `Create` flow. I argue that we're essentially "creating" coordinates here (validation is the same, required fields)

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
BenasB: This comment makes more sense, since `PollutedLocationResponse` has somewhat the same meaning as `PollutedLocationGetResponse`. I use `Get` here not as in a HTTP GET, but more as a "Retrieve"

Apsitvarkom.DataAccess/IRepository.cs
BenasB: Some food for thought about designing RESTful APIs (what we're doing here):
https://stackoverflow.com/a/28951049/13135665
https://softwareengineering.stackexchange.com/a/314069

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
BenasB: You're misunderstanding `CreatedAtAction` a bit. It tells the user that the resource they requested was created *and can be retrieved by a certain action*, in our case that action is `GetById` and we also signal that the user would need to call that action with the `response.Id` argument value. This can be seen on swagger:

image

Apsitvarkom.IntegrationTests/DataAccess/DatabaseTests.cs
BenasB: Happy path is handled in Unit tests, and this scenario could only be tested with an in-memory db, so that's why it's there. We could have more tests in the future (where validation fails, etc.), but I would prefer if we did it with another task, maybe someone else from the team wants to chip in on this

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
BenasB: Not sure what you mean here, what is "this" that AutoMapper handles?

Apsitvarkom.DataAccess/IRepository.cs
BenasB: Maybe you're right, I initially thought that maybe the `InsertAsync` could mutate the model somehow, but on the other hand I don't think it should be ever do that, the point of this method is to insert the model to the database.

BenasB: The first comment still stands as why we're returning the created model from the endpoint, although we shouldn't change the model in this repository method (I will make necessary changes).

BenasB: See this commit https://github.com/vu-vibedosa/apsitvarkom-be/pull/101/commits/021b1009b57538ec13234860f2be60f4b231affe

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
BenasB: See the convo about `InsertAsync`

Apsitvarkom.IntegrationTests/DataAccess/DatabaseTests.cs
6ar8nas: Fair.

Apsitvarkom.DataAccess/IRepository.cs
6ar8nas: Yup. Meant exactly that. We shouldn't return it from the repository insert method, but as a CRUD Create action - sure. This would also allow us to add it to the polluted locations list on the front-end without having to get all locations again I believe.

6ar8nas: The first link had a nice reference though :D Saved it to my bookmarks.

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
6ar8nas: Yeah, interpreted this wrong. My bad.

Also, that's cool. Didn't know about it.

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
6ar8nas: image
previously we had the same response type as we have now, but returned Ok with unmapped locations of type `PollutedLocation`. Is the mapping really needed, as I believe automapping takes place from when we pass `PollutedLocation`s and receive `PollutedLocationResponse`s

BenasB: No, mapping does not happen here automatically. It's just that we didn't notice because both of those objects have same properties (as of now)

BenasB: If I do this:
image
Then
image

*Note the types*

Apsitvarkom.IntegrationTests/DataAccess/DatabaseTests.cs
```diff
@@ -73,4 +73,17 @@ public async Task GetByPropertyTest()
...
+    {
+        // Try to insert the same row that was inserted in [SetUp]
+        var dbRow = DbInitializer.FakePollutedLocations.Value.Skip(3).Take(1).Single();
+
+        // Use a clean instance of the context to run the test
+        await using var context = new PollutedLocationContext(_options);
+        var dbRepository = new PollutedLocationDatabaseRepository(context);
+
+        Assert.ThrowsAsync<ArgumentException>(() => _ = dbRepository.InsertAsync(dbRow));
```
6ar8nas:
```suggestion
        Assert.ThrowsAsync<ArgumentException>(() => dbRepository.InsertAsync(dbRow));
```

I guess that's not a mistake, but any reason for this?

6ar8nas (approved):
Looks great.

Apsitvarkom.IntegrationTests/DataAccess/DatabaseTests.cs
BenasB: It's usually used to tell the reader of the code that the result of this method is discarded/not needed. I usually keep these, but since we don't return anything from `InsertAsync` we can remove it

PR #103, opened 2022-11-18
No issue: Several additions to restructuration changes

6ar8nas:
## What was done

- Some typos fixed
- Private variable naming pattern introduced, as previously we've had both m_name and _name.
- Added some regions for testing, as this seems to be a commonly used practice in our repository to make each test category stand out more.


BenasB (approved):
LGTM

PR #104, opened 2022-11-19
No issue: Clean-up after laboratory work #2

6ar8nas:
## What was done

- Logger and its related changes removed


BenasB (approved):
Thanks

PR #107, opened 2022-11-20
Extend PollutedLocation creation with title generation using Geocoder

### LINKED ISSUE #105: Extend PollutedLocation creation with title generation using Geocoder
## Steps to take

1. Add a title property to Location (comment written for its location).
2. Extend the #101 contents to now use the geocoder.
3. Adjust the geocoder configuration if needed.
4. Adjust the tests.

## Acceptance criteria

When creating a polluted location using swagger, a title is returned matching the created coordinates.

6ar8nas:
## What was done

- Moved mapping to its separate project, so that no circular dependency appears
- Added title properties to PollutedLocation and its response
- Updated the DbInitializer with Title property values.
- Created a custom PollutedLocationTitleResolver, that requests the title from the geocoder.
- Covered the implementation in preliminary tests.

Notes:
- Might need to `docker-compose` with `down -v` to test this PR's contents.
- Might need to check if we take the best possible google geocoder response result.


BenasB:
It seems like we're trying to use automapper in a not so intended way (async call when mapping). I think it's best to just ignore those properties when mapping and set the in the action itself, see PRs:
- #108 
- vu-vibedosa/apsitvarkom-fe#35

BenasB (approved):

PR #110, opened 2022-11-20
Extend IDTORepository and PollutedLocationController with Delete

### LINKED ISSUE #90: Extend IDTORepository and PollutedLocationController with Delete

6ar8nas:
## What was done

- Extended IDTORepository with Delete
- Create a `PollutedLocationIdentifyRequest`. Changed `GetById` implementation and tests to now use this newly created public model.
- Implemented the delete action in the `PollutedLocationController` at endpoint `...\Delete\`
- Readjusted the order of the `PollutedLocationResponse` properties to now return location id first.
- Covered the implementation with unit and integration tests.

Note. Didn't create a map to PollutedLocation as I don't believe it would ever be necessary - we only need the Id property and nothing else, which we can simply validate using a created validator.

Closes #90


6ar8nas:
Quick note: I found that there was no default standard for delete actions, so I've decided that we could include the location inside the response so that if we delete a location in FE, we can immediately remove the card as it is deleted real-time without having the need to load all locations again.

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
```diff
@@ -131,4 +137,32 @@ public async Task<ActionResult<PollutedLocationResponse>> Create(PollutedLocatio
...
+        var validationResult = await _pollutedLocationIdentifyValidator.ValidateAsync(pollutedLocationIdentifyRequest);
+        if (!validationResult.IsValid) return BadRequest(validationResult.Errors.Select(e => e.ErrorMessage).ToList());
+
+        try
+        {
+            var location = await _repository.GetByPropertyAsync(x => x.Id == pollutedLocationIdentifyRequest.Id);
+            if (location is null) return NotFound($"Polluted location with the specified id '{pollutedLocationIdentifyRequest.Id}' was not found.");
+
+            var mappedLocation = _mapper.Map<PollutedLocationResponse>(location);
+            if (mappedLocation is null) return StatusCode(StatusCodes.Status500InternalServerError);
```
BenasB: should these two be in the try catch block?

BenasB (approved):

PR #114, opened 2022-11-25
Introduce CleaningEvent to the codebase.

### LINKED ISSUE #113: Create TidyingEvent model, its mapping profile, dbContext, repository and controller.
## Description

As was discussed in #78, we are going to start working on the events which users can host to tidy up the polluted locations. This issue is supposed to initiate work by creating the development frame for PollutedLocations tidying up events - `TidyingEvent`s.

## Steps to take

1. In the `Apsitvarkom.Models` project create a new class `TidyingEvent`.
2. Have the `TidyingEvent` contain the PollutedLocation id, on which it is based, DateTime property for when the date for the event is set and an extra optional notes section.
3. Create a new `TidyingEventProfile` in `Apsitvarkom.Models/Mapping`. Have it match the implementation structure of the `PollutedLocationProfile`.
4. Add a DbSet for `TidyingEvent` class in `Apsitvarkom.DataAccess` existent `PollutedLocationContext`.
5. Create a new `TidyingEventDatabaseRepository` class in `Apsitvarkom.DataAccess` project. You can keep all the methods with the default `throw NotImplementedException` outcome for now - it will allow us to collaborate before someone finishes up the first repository method implementation.
6. Create several instances to initialize the database with in the DbInitializer.
7. Create a designated controller for `TidyingEvent` class. Implement GetAll, GetById and GetByProjectId actions.
8. Create maps and validators if needed.
9. Cover the implementation with tests.

6ar8nas:
## What was done

- Model `CleaningEvent` and some of its public models created, created a mapping profile for cleaning events.
- Added `CleaningEvent` to the `PollutedLocationContext` (to do this as advised [in most cases](https://learn.microsoft.com/en-us/ef/core/modeling/relationships) I added `List<CleaningEvent>` to the `PollutedLocation` class, which kind of also seems to make sense. All of related public models, mappings and validators adjusted too.
- Created `CleaningEventDatabaseRepository` and implemented most of its actions (update missing, as I did it based on the current repository of `PollutedLocation`s).
- Added some events to the DbInitializer. 
- Created a `CleaningEventController`, with GetById, GetAll
- Tests written
- Covered some previous technical debt - wrote more integration and unit tests, adjusted try/catch statements to only be around where we call db methods and etc.

This PR will allow more than one person to work in parallel, as the base for this model will already be up and running.

Closes #113 

Sorry for the big PR. Didn't expect to do it all in one, but one thing lead to another and this happened :D

Related [PR#42](https://github.com/vu-vibedosa/apsitvarkom-fe/pull/42) in front-end.


6ar8nas:
Not everything is correct with DbInitializer and PollutedLocationContext (probably OnModelCreating part [?]).
Will delve into it more later.

6ar8nas:
I'm missing something regarding Events being returned from database. They seem to exist in the db, but they're not returned from it. Will look into it tomorrow.

Apsitvarkom.Models/PollutedLocation.cs
```diff
@@ -32,4 +32,7 @@ public enum SeverityLevel
 
     /// <summary>Geographic location information about the record.</summary>
     public Location Location { get; set; } = new();
+
+    /// <summary>Collection of events linked to the location.</summary>
+    public List<TidyingEvent> Events { get; set; } = new();
```
6ar8nas: Apparently, DB doesn't return any data for this `Events` property by default, as it is (logically, due to unwanted data repetitions) not stored in the `PollutedLocation` table. This is why I didn't include it in the `PollutedLocationResponse`. Will play around and see if we could possibly return it anyways, as I believe it would be beneficial when listing all the events in the sidebar as well.

6ar8nas: resolved in `PollutedLocationDatabaseRepository`, where I had to specify that I'd like to include `Events` into the returned `PollutedLocation` instance(s).

Apsitvarkom.DataAccess/PollutedLocationContext.cs
```diff
@@ -6,16 +6,18 @@ namespace Apsitvarkom.DataAccess;
...
 public class PollutedLocationContext : DbContext, IPollutedLocationContext
 {
     public DbSet<PollutedLocation> PollutedLocations { get; private set; } = null!;
+    public DbSet<TidyingEvent> TidyingEvents { get; private set; } = null!;
     public DbContext Instance => this;
 
     public PollutedLocationContext(DbContextOptions<PollutedLocationContext> options) : base(options) { }
-
+    
     protected override void OnModelCreating(ModelBuilder modelBuilder)
```
6ar8nas, line 21: Didn't figure out a way to not allow creation of `TidyingEvent` if there is no such `PollutedLocation` yet.
Probably one way to achieve it could be introducing a `PollutedLocation` property inside of the `TidyingEvent` model and then specifying `modelBuilder.Entity<TidyingEvent>.Property(x => x.PollutedLocation).IsRequired();`

6ar8nas: I've already tried 
`modelBuilder.Entity<TidyingEvent>().HasOne<PollutedLocation>().WithMany(l => l.Events).HasForeignKey(e => e.PollutedLocationId).IsRequired();` - doesn't seem to work

6ar8nas: another workaround could be just checking if a PollutedLocation by the given id exists in controller, I suppose

Apsitvarkom.Models/Public/TidyingEventResponse.cs
```diff
@@ -0,0 +1,12 @@
...
+public class TidyingEventResponse
+{
+    public Guid Id { get; init; }
+
+    public Guid PollutedLocationId { get; set; }
+
+    public DateTime StartTime { get; set; }
+
+    public string? Notes { get; set; }
+}
```
6ar8nas: Didn't think there was any need for a validator here, hence there is none, which is different to what we have in most of the other public models.

BenasB: Response validators are used to document swagger (show schema), for example that the `Id` is not nullable

Apsitvarkom.Models/PollutedLocation.cs
BenasB: Yes, you have to tell EF core explicitly to include nested complex types

Apsitvarkom.DataAccess/PollutedLocationContext.cs
BenasB: > Probably one way to achieve it could be introducing a PollutedLocation property inside of the TidyingEvent model

And that's exactly what you should do IMO, as per [Fully defined relationships](https://learn.microsoft.com/en-us/ef/core/modeling/relationships?tabs=fluent-api%2Cfluent-api-simple-key%2Csimple-key#fully-defined-relationships) docs, they have the same exact setup: "Blog" is our "PollutedLocation" and "Post" is our "TidyingEvent". We have a 1 (polluted location) to N (many TidyingEvents) relation

Apsitvarkom.Api/Program.cs
```diff
@@ -33,6 +34,7 @@
 
 builder.Services
     .AddAutoMapper(typeof(PollutedLocationProfile))
+    .AddAutoMapper(typeof(TidyingEventProfile))
```
BenasB: Do you need this additional `AddAutoMapper`? I think that since both of our Profiles are in the same [assembly](https://learn.microsoft.com/en-us/dotnet/standard/assembly/), we only need to specify one.

> And then you let AutoMapper know in what **assemblies** are those profiles...

Ref: https://docs.automapper.org/en/stable/Dependency-injection.html

Apsitvarkom.DataAccess/PollutedLocationDatabaseRepository.cs
```diff
@@ -21,21 +21,25 @@ public PollutedLocationDatabaseRepository(IPollutedLocationContext context)
...
     }
 
     /// <inheritdoc />
-    public Task<PollutedLocation?> GetByPropertyAsync(Expression<Func<PollutedLocation, bool>> propertyCondition)
+    public async Task<PollutedLocation?> GetByPropertyAsync(Expression<Func<PollutedLocation, bool>> propertyCondition)
     {
-        return _context.PollutedLocations.FirstOrDefaultAsync(propertyCondition);
+        return await _context.PollutedLocations
+            .Include(l => l.Events)
+            .FirstOrDefaultAsync(propertyCondition);
```
BenasB: `async` `await` is not needed here, it just adds overhead, see https://stackoverflow.com/a/38017167/13135665

As a rule of thumb, if you can directly return a `Task`, you should. Might be worth checking other methods also

Apsitvarkom.DataAccess/TidyingEventDatabaseRepository.cs
```diff
@@ -0,0 +1,46 @@
...
+    public async Task<IEnumerable<TidyingEvent>> GetAllAsync()
+    {
+        return await _context.TidyingEvents.ToListAsync();
+    }
+
+    /// <inheritdoc />
+    public async Task<TidyingEvent?> GetByPropertyAsync(Expression<Func<TidyingEvent, bool>> propertyCondition)
+    {
+        return await _context.TidyingEvents.FirstOrDefaultAsync(propertyCondition);
+    }
```
BenasB: Here also for example (referring to comment about `async` `await`)

BenasB (changes requested):

Apsitvarkom.Models/Public/TidyingEventResponse.cs
BenasB: Nevermind, nullable is determined automatically, these validators to document swagger with **business rules**, but this model doesn't really have any right now

BenasB (approved):

PR #115, opened 2022-11-26
PollutedLocation Delete action adjustments - StatusCode204 instead of 200

### LINKED ISSUE #90: Extend IDTORepository and PollutedLocationController with Delete

6ar8nas:
## What was done

- Implemented Repository method `ExistsByPropertyAsync`
- Changed Delete action in `PollutedLocationController`
- Wrote tests to cover the implementation
- Renamed `TidyingEvent` to `CleaningEvent`


Apsitvarkom.Api/Controllers/PollutedLocationController.cs
```diff
@@ -154,30 +154,32 @@ public async Task<ActionResult<PollutedLocationResponse>> Create(PollutedLocatio
...
-        if (location is null) return NotFound($"Polluted location with the specified id '{pollutedLocationIdentifyRequest.Id}' was not found.");
-
-        var mappedLocation = _mapper.Map<PollutedLocationResponse>(location);
-        if (mappedLocation is null) return StatusCode(StatusCodes.Status500InternalServerError);
+        if (locationExists is false) return NotFound($"Polluted location with the specified id '{pollutedLocationIdentifyRequest.Id}' was not found.");
+        
+        var location = new PollutedLocation
+        {
+            Id = (Guid)pollutedLocationIdentifyRequest.Id!
+        };
```
BenasB: It would be worthwhile to change this to an automapper mapping?

Apsitvarkom.DataAccess/PollutedLocationDatabaseRepository.cs
```diff
@@ -42,6 +42,14 @@ orderby pollutedLocation.Location.Coordinates.DistanceTo(inRelationTo)
             .FirstOrDefaultAsync(propertyCondition);
     }
 
+    /// <inheritdoc />
+    public async Task<bool> ExistsByPropertyAsync(Expression<Func<PollutedLocation, bool>> propertyCondition)
+    {
+        return await _context.PollutedLocations
+            .Include(l => l.Events)
+            .AnyAsync(propertyCondition);
+    }
```
BenasB, line 51: Possible to return `Task` directly?

Apsitvarkom.DataAccess/TidyingEventDatabaseRepository.cs
```diff
@@ -30,6 +30,12 @@ public async Task<IEnumerable<TidyingEvent>> GetAllAsync()
         return await _context.TidyingEvents.FirstOrDefaultAsync(propertyCondition);
     }
 
+    /// <inheritdoc />
+    public async Task<bool> ExistsByPropertyAsync(Expression<Func<TidyingEvent, bool>> propertyCondition)
+    {
+        return await _context.TidyingEvents.AnyAsync(propertyCondition);
+    }
```
BenasB: Same here

Apsitvarkom.DataAccess/PollutedLocationDatabaseRepository.cs
BenasB: See https://github.com/vu-vibedosa/apsitvarkom-be/pull/114#discussion_r1033876775

BenasB (changes requested):

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
BenasB: I think you can just map the `Id` and then do `ForAllOtherMembers(opt => opt.Ignore());`

Apsitvarkom.DataAccess/PollutedLocationDatabaseRepository.cs
6ar8nas: I guess I will resolve all of these in #114. It would unnecessarily include `CleaningEventDatabaseRepository` into this as well [as the same is replicated there], which is kind of unrelated to this task (even though I technically changed the files due to renaming).

Apsitvarkom.DataAccess/TidyingEventDatabaseRepository.cs
6ar8nas: same comment as above, if that seems alright

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
6ar8nas: `ForAllOtherMembers` is deprecated as of AutoMapper 11 - used what was suggested in the migration guide:
image

BenasB (approved):

PR #119, opened 2022-11-27
Update action

### LINKED ISSUE #89: Extend IDTORepository and PollutedLocationController with Update
## Description

A simple instance update CRUD action is necessary so that we could use it to modify already existent polluted location instances. Let's join this task altogether with the next and implement this method straight into the controller as well. Use contents of #101 as a reference.

## Steps to take

1. Extend `IDTORepository` with `Update` method taking the PollutedLocationUpdateRequest instance as a parameter. Create their validators and mappers.
3. Consider which properties the PollutedLocationUpdateRequest would require for the change to be successful, if we should make all properties possible to be changed, what to do if such instance (for example by the given ID) doesn't exist yet or when we should re-evaluate the title for the location using geocoder.
4. Implement the newly created method in the PollutedLocationController class. Take the instance from post body.

## Acceptance criteria

You can update an instance using a post request in swagger, with the endpoint being PollutedLocation/Update. The implementation is covered with unit tests.

XxDoMaSxX:
## What was done

- Update action (controller endpoint),
- writen tests


Apsitvarkom.Api/Controllers/PollutedLocationController.cs
```diff
@@ -108,7 +111,7 @@ public async Task<ActionResult<PollutedLocationResponse>> GetById(string id)
     public async Task<ActionResult<PollutedLocationResponse>> Create(PollutedLocationCreateRequest pollutedLocationCreateRequest)
     {
         var validationResult = await _pollutedLocationValidator.ValidateAsync(pollutedLocationCreateRequest);
-        if (!validationResult.IsValid) return BadRequest(validationResult.Errors.Select(e => e.ErrorMessage).ToList());
+        if (!validationResult.IsValid) return BadRequest(error: validationResult.Errors.Select(e => e.ErrorMessage).ToList());
```
6ar8nas: this explicit named argument usage `error:` is not necessary.
while it's not a mistake, in a real-work scenario you wouldn't do it, so let's maybe revert this to what we've previously had.

6ar8nas: If it was done for your own clarity, you can check [this](https://davecallan.com/how-to-enable-parameter-and-type-inline-hints-in-visual-studio/) guide out.
the result: 
image

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
```diff
@@ -131,4 +134,31 @@ public async Task<ActionResult<PollutedLocationResponse>> Create(PollutedLocatio
 
         return CreatedAtAction(nameof(GetById), new { response.Id }, response);
     }
-}
\ No newline at end of file
+
+    [ProducesResponseType(StatusCodes.Status201Created)]
```
6ar8nas: When looking into the [status codes](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#2xx_success), 201 is known to be used when a new resource is created. Here we only edit an already existent resource -> we don't create a new one.

A [common practice](https://www.moesif.com/blog/technical/api-design/Which-HTTP-Status-Code-To-Use-For-Every-CRUD-App/#update) is using 200 (OK) or 204 (NoContent). As you seem to have implemented it as an `Ok` (line 157), let's keep it at 
```suggestion
    [ProducesResponseType(StatusCodes.Status200OK)]
```

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
```diff
@@ -131,4 +134,31 @@ public async Task<ActionResult<PollutedLocationResponse>> Create(PollutedLocatio
...
+    [ProducesResponseType(StatusCodes.Status400BadRequest, Type = typeof(List<string>))]
+    [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+    [HttpPut("Update")]
+    public async Task<ActionResult<PollutedLocationResponse>> Update(PollutedLocationUpdateRequest pollutedLocationUpdateRequest)
+    {
+        var validatorResponse = await _pollutedLocationUpdateValidator.ValidateAsync(pollutedLocationUpdateRequest);
```
6ar8nas: kind of a nitpick, but in the example just above, we used a variable named `validationResult`: keeping consistent with variable naming is useful when writing a collaborative code - in most cases (unless it's a really old code in need to be refactored) you would want to stick with the style around you. let's change that to match.

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
```diff
@@ -131,4 +134,31 @@ public async Task<ActionResult<PollutedLocationResponse>> Create(PollutedLocatio
...
+        if (!validatorResponse.IsValid) return BadRequest(validatorResponse.Errors.Select(e => e.ErrorMessage).ToList());
+
+        var existingLocation = await _repository.GetByPropertyAsync(x => x.Id == pollutedLocationUpdateRequest.Id);
```
6ar8nas: just as you've done below, with repository calls we expect that something might go wrong and surround these calls with try/catch blocks, returning `Status500InternalServerError` if something goes wrong.

6ar8nas: if done like this, you couldn't really define the `existingLocation` in the try/catch (as the variable would go out of scope as soon as the try/catch ends), so you can refer to what is done [here](https://github.com/vu-vibedosa/apsitvarkom-be/blob/staging/Apsitvarkom.Api/Controllers/PollutedLocationController.cs#L163) for an example of what can be done.

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
```diff
@@ -131,4 +134,31 @@ public async Task<ActionResult<PollutedLocationResponse>> Create(PollutedLocatio
...
+        if (existingLocation is null) return StatusCode(StatusCodes.Status500InternalServerError);
```
6ar8nas: if this location is null, it means that we couldn't acquire the location from our database, because there was no location with this id, not that something with the server was wrong.
`500InternalServerError` implies that the server encountered an unexpected condition that prevented it from fulfilling the request, and we didn't encounter any problem with the server here - we just didn't find the needed object.

Let's change this return statement to return a `Status404NotFound` instead. Please refer to messages in [`Delete`](https://github.com/vu-vibedosa/apsitvarkom-be/blob/staging/Apsitvarkom.Api/Controllers/PollutedLocationController.cs#L173) or [`GetById`](https://github.com/vu-vibedosa/apsitvarkom-be/blob/staging/Apsitvarkom.Api/Controllers/PollutedLocationController.cs#L97) (note, these are the same messages).

6ar8nas: Don't forget to add that the action call can now also produce a `NotFound` response type, as defined with other types above.

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
```diff
@@ -131,4 +134,31 @@ public async Task<ActionResult<PollutedLocationResponse>> Create(PollutedLocatio
...
+       
+        var mappedLocation = _mapper.Map<PollutedLocationUpdateRequest, PollutedLocation>(pollutedLocationUpdateRequest, existingLocation);
+        if (mappedLocation is null) return StatusCode(StatusCodes.Status500InternalServerError);
+
+        try
+        {
+            await _repository.UpdateAsync(mappedLocation);
+
+            return Ok(mappedLocation);
```
6ar8nas: Let's move out of the the try/catch. Technically it doesn't mean much difference and we'd do it just for constancy reasons (or if any post-update condition would be introduced). 
For the cases above, that acted similarly, they're already fixed in an upcoming PR of my own #114

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
```diff
@@ -131,4 +134,31 @@ public async Task<ActionResult<PollutedLocationResponse>> Create(PollutedLocatio
 
         return CreatedAtAction(nameof(GetById), new { response.Id }, response);
     }
-}
\ No newline at end of file
+
+    [ProducesResponseType(StatusCodes.Status201Created)]
+    [ProducesResponseType(StatusCodes.Status400BadRequest, Type = typeof(List<string>))]
+    [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+    [HttpPut("Update")]
```
6ar8nas: After a [quick look](https://stackoverflow.com/questions/31089221/what-is-the-difference-between-put-post-and-patch), this what we do here is more of a `HttpPatch` rather than `HttpPut`

Apsitvarkom.DataAccess/IRepository.cs
```diff
@@ -20,4 +20,6 @@ public interface IRepository<T> where T: class
     /// <param name="modelToInsert">Model to be inserted into the repository.</param>
     /// <returns>The Task to insert the model.</returns>
     public Task InsertAsync(T modelToInsert);
+
+    public Task UpdateAsync(T modelToUpdate);
```
6ar8nas, line 32: As can be seen with the interface methods above, write XML documentation for this method. That documentation can be seen when we hover over the method or over its params.
`<summary>Text</summary>` notes generally what this method does.
`<param name="variable">Text</param>` notes what kind of a parameter is expected.
`<returns>Text</returns>` notes what is returned by the method.

Apsitvarkom.DataAccess/PollutedLocationDatabaseRepository.cs
```diff
@@ -44,4 +44,10 @@ public async Task InsertAsync(PollutedLocation modelToInsert)
         await _context.PollutedLocations.AddAsync(modelToInsert);
         await _context.Instance.SaveChangesAsync();
     }
+
+    public async Task UpdateAsync(PollutedLocation modelToUpdate)
```
6ar8nas, line 61: after you write the XML documentation, in every interface implementation, you can inherit that documentation with `<inheritdoc />` as done above.

Apsitvarkom.DataAccess/IRepository.cs
6ar8nas: Note: XML comments always start with `///`, not `//`

Apsitvarkom.Models/Mapping/PollutedLocationProfile.cs
```diff
@@ -23,6 +23,9 @@ private void MapRequests()
         CreateMap<PollutedLocationCreateRequest, PollutedLocation>()
             .ForMember(x => x.Spotted, opt => opt.Ignore())
             .ForMember(x => x.Id, opt => opt.Ignore());
+        CreateMap<PollutedLocationUpdateRequest, PollutedLocation>()
+            .ForMember(x => x.Spotted, opt => opt.Ignore())
+            .ForMember(x => x.Location, opt => opt.Ignore()); ;
```
6ar8nas:
```suggestion
            .ForMember(x => x.Location, opt => opt.Ignore());
```

6ar8nas: seems like a misplaced `;`

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
6ar8nas: `mappedLocation` is of type `PollutedLocation`, however in the method signature you state, that you will return `ActionResult<PollutedLocationResponse>`. we need to map this once again for the types to match.
Take a look on how this can be achieved [here](https://github.com/vu-vibedosa/apsitvarkom-be/blob/staging/Apsitvarkom.Api/Controllers/PollutedLocationController.cs#L138)

6ar8nas: Note: Do this before the UpdateAsync call, so that in case we can't return proper results and fail, the data wouldn't be changed as wel..

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
```diff
@@ -131,4 +134,31 @@ public async Task<ActionResult<PollutedLocationResponse>> Create(PollutedLocatio
...
+    public async Task<ActionResult<PollutedLocationResponse>> Update(PollutedLocationUpdateRequest pollutedLocationUpdateRequest)
+    {
+        var validatorResponse = await _pollutedLocationUpdateValidator.ValidateAsync(pollutedLocationUpdateRequest);
+        if (!validatorResponse.IsValid) return BadRequest(validatorResponse.Errors.Select(e => e.ErrorMessage).ToList());
+
+        var existingLocation = await _repository.GetByPropertyAsync(x => x.Id == pollutedLocationUpdateRequest.Id);
+        if (existingLocation is null) return StatusCode(StatusCodes.Status500InternalServerError);
+       
+        var mappedLocation = _mapper.Map<PollutedLocationUpdateRequest, PollutedLocation>(pollutedLocationUpdateRequest, existingLocation);
```
6ar8nas: this is wrong. if we send all the properties, it works as expected, but if for example we would like to only update the severity, radius, progress and notes would become the default values of those types (string? -> null, int -> 0), instead of keeping the previous values.
We should remove this mapping and instead put only the non-null variables to the recently extracted PollutedLocation instance.

6ar8nas (changes requested):
Didn't look into tests as of yet, as changes are needed.

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
6ar8nas: changed it myself when resolving the merge conflict, hence closing the comment.

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
XxDoMaSxX: should I check each values separetely and update only those which were changed ?
like : 
`if (instanceFromdb.field1 != instanceToUpdate.field1) instanceFromdb.field1 = instanceToUpdate.field1 ...`

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
XxDoMaSxX: I guess for future usage it would be better if I return full PollutedObject, so I should change method signature. what do you think bout that ?

6ar8nas: No, you should stick to the `PollutedLocationResponse` - `PollutedLocation` is a business-logic object, that we on purpose don't expose to be used in the actions.

Apsitvarkom.Api/Controllers/PollutedLocationController.cs
6ar8nas: No, I don't think that it is necessary. First of all, even if the data is the same, the re-assignment with the same value is not an expensive operation, which doesn't make any difference to the end result. Secondly, update request variables can be null (meaning they're optional) - this would allow the user to not specify the value at all if they wanted to keep the value the same.

6ar8nas (approved):

PR #43, opened 2022-10-04
Make PollutedLocationDTOFileRepository FromContent string parameter optional

XxDoMaSxX:
## What was done

- test
- -function parameter default value

PR #122, opened 2022-11-30
Internationalization for the title of PollutedLocation model

### LINKED ISSUE #121: Internationalization for the title of PollutedLocation model
## Description

After the implementation of [apsitvarkom-fe/#44](https://github.com/vu-vibedosa/apsitvarkom-fe/issues/44), let's adapt to this change and have titles translated into several languages.

## Steps to take

1. Change Geocoder to now make as many requests, as the count of languages we support.
2. Adjust the response models accordingly.
3. Write tests for the implementation.

## Acceptance criteria

Titles are now returned in multiple languages.

6ar8nas:
## What was done

- Changed Geocoder to now make as many requests, as the count of languages we support.
- Adjusted the response models accordingly.
- Wrote tests for the implementation.

Closes #121 

Related PR in front-end: [#43](https://github.com/vu-vibedosa/apsitvarkom-fe/pull/48)


Apsitvarkom.Models/Public/LocationTitleResponse.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    public LocationCode Code { get; set; }
+    public string Name { get; set; } = string.Empty;
+}
+
+public class LocationTitleResponseValidator : AbstractValidator<LocationTitleResponse>
+{
+    public LocationTitleResponseValidator()
+    {
+        RuleFor(x => x.Name).NotNull();
```
6ar8nas: Swagger showed Name as nullable, so I added this. However now it shows it as required, which is a unique case among all the response models, so I'm not too sure if this is needed.

Apsitvarkom.DataAccess/GoogleGeocoder.cs
```diff
@@ -22,13 +24,17 @@ public GoogleGeocoder(HttpClient httpClient, IApiKeyProvider apiKeyProvider)
         _httpClient = httpClient;
     }
 
-    public async Task<string?> ReverseGeocodeAsync(Coordinates coordinates)
+    public async Task<List<LocationTitle>> ReverseGeocodeAsync(Coordinates coordinates)
```
6ar8nas: this method is kind of a Frankenstein now and doesn't really comply with single responsibility principle. will maybe look into fixing it.

6ar8nas:
Several integration tests are failing; however those fails are solely due to testing in-memory.

BenasB (changes requested):
So the current returned response is this:
image

let's see why it's not that great.

So right now it's something of a type `List<(LocationCode, string)>`.

Is this object valid then?
```json
[
  {
    "code": "en"
    "name": "title in english"
  },
  {
    "code": "lt"
    "name": "title in lithuanian"
  },
  {
    "code": "lt"
    "name": "another one"
  },
]
```

It certainly is a `List<(LocationCode, string)>`, although we will *try* not to return values like that (but there's nothing restricting us from doing so). The swagger schema also indicates the same problem:

image

Another one:
```json
[
  {
    "code": "en"
    "name": "title in english"
  }
]
```

Or:
```json
[]
```

The problem here is that we can only "tell" which objects are in a way we want them to be during **runtime** (like you do the check in the validator). Ideally, we would want such a thing be known at **compile** time (e.g. if we are creating a translated title, we include all languages). So we could start off with something like:

```csharp
public class TranslatedString { // Ideally this can represent any translated string, no need to use is strictly for location title
  public string English { get; set; }
  public string Lithuanian { get; set; }
}
```

or something more generic would be even better (not sure how well this goes with EF core):

```csharp
public class Translated<T> {
  public T English { get; set; }
  public T Lithuanian { get; set; }
}
```

and then in `Location` have 

```csharp
public class Location
{
    public TranslatedString Title { get; set; } = new();
    public Coordinates Coordinates { get; set; } = new();
}
```
then when you are creating the `Title` field, the language fields must be initialized. And this way we are sure at **compile time** that these fields will be there. Also swagger now tells a different story (of course you would probably want to have a `TranslatedStringResponse` with `en` instead of `english` field):

image

Also compare the example values in swagger:

Before:
image

After:
image

The second option makes more sense IMO.

There's one more thing that we need – the ability to iterate through the languages (for example in geocoder we want to iterate through the languages and retrieve a value for each of them), and we can't do it right now because the only place where we define the languages are in the property names (which we can't iterate through (at least not without reflections)). One way to solve this would be:

```csharp
public enum SupportedLanguages {
  English, Lithuanian
}

public class TranslatedString {
  public string English { get; private set; }
  public string Lithuanian { get; private set; }

  public void Update(SupportedLanguages language, string value)
    {
        _ = language switch
        {
            SupportedLanguages.English => English = value,
            SupportedLanguages.Lithuanian => Lithuanian = value,
            _ => throw new InvalidOperationException()
        };
    }
}
```

This means we need to retype the supported languages twice: once in the enum and once in the properties, however I think this is definitely an easy tradeoff for compile time safety. Notice that properties now have `private set` and we should update the `TranslatedString` values using the `Update` method. So Geocoder part can now look something like this:

Have something like this:
```csharp
public static string GetSupportedLocale(SupportedLanguages language) =>
    language switch
    {
        SupportedLanguages.English => "en",
        SupportedLanguages.Lithuanian => "lt",
        _ => "en"
    }
```
And then in geocoder have 2 `ReverseGeocodeAsync` overloads where one can be used to get a single language and the other gets all of the languages from `TranslatedString`:
```csharp
    public async Task<string?> ReverseGeocodeAsync(Coordinates coordinates, SupportedLanguages language)
    {
        var query = $"json?latlng={coordinates.Latitude.ToString(CultureInfo.InvariantCulture)},{coordinates.Longitude.ToString(CultureInfo.InvariantCulture)}&language={TranslatedString.GetSupportedLocale(language)}&key={_apiKey}";
        // query the api
    }

    public async Task<TranslatedString> ReverseGeocodeAsync(Coordinates coordinates)
    {
        var title = new TranslatedString();
        foreach (var language in (SupportedLanguages[]) Enum.GetValues(typeof(SupportedLanguages)))
        {
            // look into Task.WhenAll here
            title.Update(language, await ReverseGeocodeAsync(coordinates, language) ?? string.Empty);
        }
        return title;
    }
```

All in all I think this (or similar) is the way to go. On the front end it will also be easier to just have a concrete type instead of a list that would need checking during runtime.

Apsitvarkom.DataAccess/GoogleGeocoder.cs
```diff
@@ -22,15 +22,29 @@ public GoogleGeocoder(HttpClient httpClient, IApiKeyProvider apiKeyProvider)
...
 
+    public async Task<Translated<string>> ReverseGeocodeAsync(Coordinates coordinates)
+    {
+        var title = new Translated<string>();
+        var languages = (SupportedLanguages[])Enum.GetValues(typeof(SupportedLanguages));
+
+        var tasks = languages.Select(language => ReverseGeocodeAsync(coordinates, language));
+        var geocodeResults = await Task.WhenAll(tasks);
+
+        var pairs = geocodeResults.Zip(languages, (name, language) => new { Language = language, Name = name ?? string.Empty });
```
BenasB: You could make the `Zip` before awaiting tasks (have them be a pair of `(SupportedLanguage, Task<string?>)` instead of `(SupportedLanguage,  string?)`) because we're now expecting the tasks to be in the same order of the languages in `SupportedLanguages`

Apsitvarkom.Models/Public/TranslatedResponse.cs
```diff
@@ -0,0 +1,18 @@
+﻿using FluentValidation;
+
+namespace Apsitvarkom.Models.Public;
+
+public class TranslatedResponse<T>
+{
+    public T en { get; set; } = default!;
+    public T lt { get; set; } = default!;
```
BenasB: Having these in PascalCase will still return them as camelCase, so maybe it's worth keeping consistency and change these to PascalCase?

BenasB (approved):
Some comments left, but overall looks nice now

Issue #126, opened 2022-12-01
Introduce some changes to the CleaningEvent


6ar8nas:
## Description

The initially implemented `CleaningEvent` model could use an upgrade: 
- users would likely welcome the ability to filter out events which have already ended from those that hadn't yet happened.

## Steps to take

1. Adjust the model introducing the above-defined requests.

## Acceptance criteria

`CleaningEventResponse` now includes a value for whether the event has already ended.
A controller update action endpoint is created specifically for updating the event after it's finished.

PR #130, opened 2022-12-02
Add value check constraints for database columns

### LINKED ISSUE #129: Add value check constraints for database columns
## Description

In the way we currently access and use the database after validation of the data to be inserted, it will always contain only requirements-compliant data. However, in a real-world scenario, we would ideally want to assure that we wouldn't be able to insert wrong data straight into the database regardless of our API actions, thus we would need some check constraints.

## Steps to take

1. Read into [Check constraints in EF Core](https://learn.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.relationalentitytypebuilderextensions.hascheckconstraint?view=efcore-6.0) material.
2. Figure out what kind of column constraints we would need (navigating through already existent validators would help here).
3. Add those constraints using `modelBuilder` in `PollutedLocationContext`'s method `OnModelCreating` 
4. Cover the constraints using in-memory testing.

## Acceptance criteria

Check constraints added and visible in adminer.
Integration tests, showing failing inserts with invalid data written.

6ar8nas:
## What was done

- Changed the way we store Enum values in the db to the [default](https://dev.mysql.com/doc/refman/8.0/en/enum.html#:~:text=An%20ENUM%20is%20a%20string,specification%20at%20table%20creation%20time.) SQL behavior - using strings.
- Added check constraints for all PollutedLocation values. These are to be expanded with #126 

Integration testing, as was described in the issue, is unavailable using in-memory testing. Tested the behavior manually - everything seems to work fine.

Closes #129

### DIFF
--- a/Apsitvarkom.DataAccess/PollutedLocationContext.cs
+++ b/Apsitvarkom.DataAccess/PollutedLocationContext.cs
@@ -27,7 +27,7 @@ public class PollutedLocationContext : DbContext, IPollutedLocationContext
         modelBuilder.Entity<PollutedLocation>()
             .HasCheckConstraint("CK_PollutedLocation_Radius", "\"Radius\" >= 1")
             .HasCheckConstraint("CK_PollutedLocation_Progress", "\"Progress\" >= 0 and \"Progress\" <= 100")
-            .HasCheckConstraint("CK_PollutedLocation_Severity", $"\"Severity\" in ({string.Join(", ", ((SupportedLanguages[])Enum.GetValues(typeof(SupportedLanguages))).Select(p => $"`{p}'"))})")
+            .HasCheckConstraint("CK_PollutedLocation_Severity", $"\"Severity\" in ({string.Join(", ", ((SupportedLanguages[])Enum.GetValues(typeof(SupportedLanguages))).Select(p => $"'{p}'"))})")
             .OwnsOne(l => l.Location)
             .OwnsOne(l => l.Coordinates)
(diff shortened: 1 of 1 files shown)


BaVilius (approved):
Looks good to me as far as grammar and cleanliness of the code goes. However I am not familiar with the ModelBuilder class.

Apsitvarkom.DataAccess/PollutedLocationContext.cs
```diff
@@ -20,8 +20,17 @@ public class PollutedLocationContext : DbContext, IPollutedLocationContext
...
-        // Reference: https://learn.microsoft.com/en-us/ef/core/modeling/owned-entities
-        modelBuilder.Entity<PollutedLocation>().OwnsOne(l => l.Location).OwnsOne(l => l.Coordinates);
+        modelBuilder.Entity<PollutedLocation>().Property(x => x.Severity).HasConversion(
+            v => v.ToString(),
+            v => (PollutedLocation.SeverityLevel)Enum.Parse(typeof(PollutedLocation.SeverityLevel), v)
+        );
+        modelBuilder.Entity<PollutedLocation>()
+            .HasCheckConstraint("CK_PollutedLocation_Radius", "\"Radius\" >= 1")
+            .HasCheckConstraint("CK_PollutedLocation_Progress", "\"Progress\" >= 0 and \"Progress\" <= 100")
+            .HasCheckConstraint("CK_PollutedLocation_Severity", "\"Severity\" in ('Low', 'Moderate', 'High')")
```
BenasB: > `'Low', 'Moderate', 'High'`

You can probably get these values dynamically from the enum, right? We should not hard code them if we can

BenasB (changes requested):

BenasB (approved):

PR #131, opened 2022-12-03
Implement Create for CleaningEvent

### LINKED ISSUE #128: Add Create endpoint for the CleaningEventController
## Description

Implement a new `Create` method for the `CleaningEventController`, which should take the `CleaningEventCreateRequest` as a `Post` body.

## Steps to take

1. Take a look at the create implementation found in the `PollutedLocationController`. You can use it as a reference. 
2. Create the needed public models, its validators, adjust the cleaning event mapping profile.
3. Write unit tests to cover the implemented functionality in the controller, validator, mapper and other parts if modified.

## Acceptance criteria

You can create an instance of CleaningEvent using a post request in swagger, with the endpoint /Create/.
The implementation is covered with Unit tests.

XxDoMaSxX:
## What was done

-Create endpoint in CleaningEventController 
-tests written

### DIFF
--- a/Apsitvarkom.Api/Controllers/CleaningEventController.cs
+++ b/Apsitvarkom.Api/Controllers/CleaningEventController.cs
@@ -14,18 +14,18 @@ public class CleaningEventController : ControllerBase
     private readonly IRepository<CleaningEvent> _repository;
     private readonly IMapper _mapper;
     private readonly IValidator<ObjectIdentifyRequest> _objectIdentifyValidator;
-    private readonly IValidator<CleaningEventCreateRequest> _cleaningEventCreateRequestValdiator;
+    private readonly IValidator<CleaningEventCreateRequest> _cleaningEventCreateRequestValidator;
 
     public CleaningEventController(
         IRepository<CleaningEvent> repository, 
         IMapper mapper, 
         IValidator<ObjectIdentifyRequest> objectIdentifyValidator,
-        IValidator<CleaningEventCreateRequest> cleaningEventCreateRequestValdiator)
+        IValidator<CleaningEventCreateRequest> cleaningEventCreateRequestValidator)
     {
         _repository = repository;
         _mapper = mapper;
         _objectIdentifyValidator = objectIdentifyValidator;
-        _cleaningEventCreateRequestValdiator = cleaningEventCreateRequestValdiator;
+        _cleaningEventCreateRequestValidator = cleaningEventCreateRequestValidator;
     }
 
     [ProducesResponseType(StatusCodes.Status200OK)]
@@ -85,7 +85,7 @@ public class CleaningEventController : ControllerBase
     [HttpPost("Create")]
     public async Task<ActionResult<CleaningEventResponse>> Create(CleaningEventCreateRequest cleaningEventCreateRequest)
     {
-        var validationResult = await _cleaningEventCreateRequestValdiator.ValidateAsync(cleaningEventCreateRequest);
+        var validationResult = await _cleaningEventCreateRequestValidator.ValidateAsync(cleaningEventCreateRequest);
         if (!validationResult.IsValid) return BadRequest(validationResult.Errors.Select(e => e.ErrorMessage).ToList());
 
         var cleaningEventDefaults = new CleaningEvent
--- a/Apsitvarkom.Api/Controllers/CleaningEventController.cs
+++ b/Apsitvarkom.Api/Controllers/CleaningEventController.cs
@@ -14,18 +14,18 @@ public class CleaningEventController : ControllerBase
     private readonly IRepository<CleaningEvent> _repository;
     private readonly IMapper _mapper;
     private readonly IValidator<ObjectIdentifyRequest> _objectIdentifyValidator;
-    private readonly IValidator<CleaningEventCreateRequest> _cleaningEventCreateRequestValidator;
+    private readonly IValidator<CleaningEventCreateRequest> _cleaningEventCreateValidator;
 
     public CleaningEventController(
         IRepository<CleaningEvent> repository, 
         IMapper mapper, 
         IValidator<ObjectIdentifyRequest> objectIdentifyValidator,
-        IValidator<CleaningEventCreateRequest> cleaningEventCreateRequestValidator)
+        IValidator<CleaningEventCreateRequest> cleaningEventCreateValidator)
     {
         _repository = repository;
         _mapper = mapper;
         _objectIdentifyValidator = objectIdentifyValidator;
-        _cleaningEventCreateRequestValidator = cleaningEventCreateRequestValidator;
+        _cleaningEventCreateValidator = cleaningEventCreateValidator;
     }
 
     [ProducesResponseType(StatusCodes.Status200OK)]
@@ -85,7 +85,7 @@ public class CleaningEventController : ControllerBase
     [HttpPost("Create")]
     public async Task<ActionResult<CleaningEventResponse>> Create(CleaningEventCreateRequest cleaningEventCreateRequest)
     {
-        var validationResult = await _cleaningEventCreateRequestValidator.ValidateAsync(cleaningEventCreateRequest);
+        var validationResult = await _cleaningEventCreateValidator.ValidateAsync(cleaningEventCreateRequest);
         if (!validationResult.IsValid) return BadRequest(validationResult.Errors.Select(e => e.ErrorMessage).ToList());
 
         var cleaningEventDefaults = new CleaningEvent
--- a/Apsitvarkom.Api/Controllers/CleaningEventController.cs
+++ b/Apsitvarkom.Api/Controllers/CleaningEventController.cs
@@ -14,15 +14,18 @@ public class CleaningEventController : ControllerBase
     private readonly IRepository<CleaningEvent> _repository;
     private readonly IMapper _mapper;
     private readonly IValidator<ObjectIdentifyRequest> _objectIdentifyValidator;
+    private readonly IValidator<CleaningEventCreateRequest> _cleaningEventCreateRequestValdiator;
 
     public CleaningEventController(
         IRepository<CleaningEvent> repository, 
         IMapper mapper, 
-        IValidator<ObjectIdentifyRequest> objectIdentifyValidator)
+        IValidator<ObjectIdentifyRequest> objectIdentifyValidator,
+        IValidator<CleaningEventCreateRequest> cleaningEventCreateRequestValdiator)
     {
         _repository = repository;
         _mapper = mapper;
         _objectIdentifyValidator = objectIdentifyValidator;
+        _cleaningEventCreateRequestValdiator = cleaningEventCreateRequestValdiator;
     }
 
     [ProducesResponseType(StatusCodes.Status200OK)]
@@ -72,5 +75,37 @@ public class CleaningEventController : ControllerBase
         if (mappedEvent is null) return StatusCode(StatusCodes.Status500InternalServerError);
 
         return Ok(mappedEvent);
+    }
+
+    [ProducesResponseType(StatusCodes.Status201Created)]
+    [ProducesResponseType(StatusCodes.Status400BadRequest, Type = typeof(List<string>))]
+    [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+    [HttpPost("Create")]
+    public async Task<ActionResult<CleaningEventResponse>> Create(CleaningEventCreateRequest cleaningEventCreateRequest)
+    {
+        var validationResult = await _cleaningEventCreateRequestValdiator.ValidateAsync(cleaningEventCreateRequest);
+        if (!validationResult.IsValid) return BadRequest(validationResult.Errors.Select(e => e.ErrorMessage).ToList());
+
+        var cleaningEventDefaults = new CleaningEvent
+        {
+            Id = Guid.NewGuid()
+        };
+
+        var mappedCleaningEvent = _mapper.Map<CleaningEventCreateRequest, CleaningEvent>(cleaningEventCreateRequest, cleaningEventDefaults);
+        if (mappedCleaningEvent is null) return StatusCode(StatusCodes.Status500InternalServerError);
+
+        var response = _mapper.Map<CleaningEventResponse>(mappedCleaningEvent);
+        if (response is null) return StatusCode(StatusCodes.Status500InternalServerError);
+
+        try
+        {
+            await _repository.InsertAsync(mappedCleaningEvent);
+        }
+        catch (Exception)
+        {
+            return StatusCode(StatusCodes.Status500InternalServerError);
+        }
+
+        return CreatedAtAction(nameof(GetById), new { response.Id }, response);
     }
 }
--- a/Apsitvarkom.Api/Controllers/CleaningEventController.cs
+++ b/Apsitvarkom.Api/Controllers/CleaningEventController.cs
@@ -12,6 +12,7 @@ namespace Apsitvarkom.Api.Controllers;
 public class CleaningEventController : ControllerBase
 {
     private readonly IRepository<CleaningEvent> _repository;
+    private readonly IPollutedLocationRepository _pollutedLocationRepository;
     private readonly IMapper _mapper;
     private readonly IValidator<ObjectIdentifyRequest> _objectIdentifyValidator;
     private readonly IValidator<CleaningEventCreateRequest> _cleaningEventCreateValidator;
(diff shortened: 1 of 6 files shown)


Apsitvarkom.Api/Controllers/CleaningEventController.cs
```diff
@@ -14,15 +14,18 @@ public class CleaningEventController : ControllerBase
     private readonly IRepository<CleaningEvent> _repository;
     private readonly IMapper _mapper;
     private readonly IValidator<ObjectIdentifyRequest> _objectIdentifyValidator;
+    private readonly IValidator<CleaningEventCreateRequest> _cleaningEventCreateRequestValdiator;
```
6ar8nas: you made a mistake here writing the word `Validator`.
also looking at the validator name above, we could omit the `Create` part and instead have the name as follows:
```suggestion
    private readonly IValidator<CleaningEventCreateRequest> _cleaningEventCreateValidator;
```

Apsitvarkom.Api/Controllers/CleaningEventController.cs
```diff
@@ -14,15 +14,18 @@ public class CleaningEventController : ControllerBase
...
 
     public CleaningEventController(
         IRepository<CleaningEvent> repository, 
         IMapper mapper, 
-        IValidator<ObjectIdentifyRequest> objectIdentifyValidator)
+        IValidator<ObjectIdentifyRequest> objectIdentifyValidator,
+        IValidator<CleaningEventCreateRequest> cleaningEventCreateRequestValdiator)
```
6ar8nas:
```suggestion
        IValidator<CleaningEventCreateRequest> cleaningEventCreateValidator)
```

Apsitvarkom.Models/Public/CleaningEventCreateRequest.cs
```diff
@@ -0,0 +1,23 @@
+﻿using Apsitvarkom.Models.Public;
```
6ar8nas: this reference is not used

Apsitvarkom.Models/Public/CleaningEventCreateRequest.cs
```diff
@@ -0,0 +1,23 @@
...
+
+public class CleaningEventCreateRequest
+{
+    public Guid? PollutedLocationId { get; set; }
+
+    public DateTime? StartTime { get; set; }
+
+    public string? Notes { get; set; }
+}
+
```
6ar8nas, line 13:
```suggestion
```

Apsitvarkom.Models/Public/CleaningEventCreateRequest.cs
```diff
@@ -0,0 +1,23 @@
...
+
+public class CleaningEventCreateRequestValidator : AbstractValidator<CleaningEventCreateRequest>
+{
+    public CleaningEventCreateRequestValidator()
+    {
+        RuleFor(l => l.StartTime).NotNull();
```
6ar8nas: I believe we should also check if this DateTime is not a date in the past, as this wouldn't make sense. You can check that with `GreaterThan` method.

Apsitvarkom.Api/Controllers/CleaningEventController.cs
```diff
@@ -73,4 +76,36 @@ public async Task<ActionResult<CleaningEventResponse>> GetById([FromQuery] Objec
 
         return Ok(mappedEvent);
     }
+
+    [ProducesResponseType(StatusCodes.Status201Created)]
+    [ProducesResponseType(StatusCodes.Status400BadRequest, Type = typeof(List<string>))]
+    [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+    [HttpPost("Create")]
+    public async Task<ActionResult<CleaningEventResponse>> Create(CleaningEventCreateRequest cleaningEventCreateRequest)
```
6ar8nas, line 93: There is an issue with this. If the polluted location id doesn't exist, we return a StatusCode500, which means there has been an internal server error and there was no such internal issue - we should instead return an error code 404 with an error message `Polluted location with the specified id '{cleaningEventCreateRequest.PollutedLocationId}' was not found.`

This can be probably achieved in two ways:
- Checking what kind of exception we get - I believe the `InnerException` in the try/catch of the `InsertAsync` holds a dictionary with `ConstraintName` key, whose value is FK_CleaningEvents_PollutedLocations_PollutedLocationId. If we ever find out, that this is the exact circumstances of the exception that has happened, we would return a different status code.
- Adding a new method to `IRepository` interface `ParentExistsByPropertyAsync`, which would have one purpose, defined by its name. In instances, that do not have any parent (such as the `PollutedLocation`), we would simply throw `NotSupportedException`.

I personally think that the second approach is a better way to go, but I could be wrong, so perhaps we could check with @BenasB about this.

6ar8nas (changes requested):

6ar8nas:
Closes #128

6ar8nas:
Don't forget to link the issue you're resolving to the PR next time.

6ar8nas:
Solved merge conflicts. Next time please pull the most recent changes before starting to work on an issue and merge your pull requests once they're ready so that others could already use the changes that you've done.

Apsitvarkom.UnitTests/Models/Validation/CleaningEventCreateRequestValidationTests.cs
```diff
@@ -0,0 +1,63 @@
+﻿using Apsitvarkom.Models;
```
6ar8nas: this reference is not used.

Apsitvarkom.UnitTests/Models/Validation/CleaningEventCreateRequestValidationTests.cs
```diff
@@ -0,0 +1,63 @@
...
+using Apsitvarkom.Models.Public;
+using FluentValidation;
+
+namespace Apsitvarkom.UnitTests.Models.Validation;
+public class CleaningEventCreateRequestValidationTests
```
6ar8nas:
```suggestion
namespace Apsitvarkom.UnitTests.Models.Validation;

public class CleaningEventCreateRequestValidationTests
```

Apsitvarkom.UnitTests/Api/Controllers/CleaningEventControllerTests.cs
```diff
@@ -41,6 +41,44 @@ public class CleaningEventControllerTests
         }
     };
 
+    private readonly IEnumerable<PollutedLocation> PollutedLocations = new List<PollutedLocation>
```
6ar8nas: you only use this variable for `pollutedLocationId`s. using these `Id`s is as random as creating a random guid using `Guid.NewGuid()` and we wouldn't need to have this big list.

Apsitvarkom.UnitTests/Api/Controllers/CleaningEventControllerTests.cs
```diff
@@ -214,6 +254,83 @@ public async Task GetById_RepositoryThrows_Status500InternalServerErrorReturned(
...
+        Assert.Multiple(() =>
+        {
+            Assert.That(resultLocation.PollutedLocationId, expression: Is.EqualTo(createRequest.PollutedLocationId));
+            Assert.That(resultLocation.StartTime, Is.EqualTo(createRequest.StartTime));
+            Assert.That(resultLocation.Notes, Is.EqualTo(createRequest.Notes));
+        });
+    }
+
+    [Test]
+    public async Task Create_RepositoryThrows_NotExistingPollutedLocationIdProvided_Status500InternalServerErrorReturned()
```
6ar8nas:
```suggestion
    public async Task Create_RepositoryThrows_Status500InternalServerErrorReturned()
```
the error you're trying to replicate doesn't necessarily originate from pollutedLocationId not existing in the db. Suggesting this change not only because of that, but as was described in the previous review, where I touched on the implementation and not tests, I've specified, that it is not proper behavior that we have there currently.

Apsitvarkom.UnitTests/Models/Mapping/CleaningEventMappingTests.cs
```diff
@@ -58,5 +58,33 @@ public void CleaningEventToCleaningEventResponse()
             Assert.That(response.PollutedLocationId, Is.EqualTo(businessLogicObject.PollutedLocationId));
         });
     }
+
+    [Test]
+    public void CleaningEventCreateRequestToCleaningEvent()
```
6ar8nas: you put it in the `Response mappings` region, while this is a request -> business object mapping.

Apsitvarkom.UnitTests/Models/Mapping/CleaningEventMappingTests.cs
```diff
@@ -58,5 +58,33 @@ public void CleaningEventToCleaningEventResponse()
...
+
+        var createModel = new CleaningEventCreateRequest
+        {
+            StartTime = new DateTime(2022, 9, 16, 21, 43, 31).ToUniversalTime(),
+            Notes = "Hello world",
+            PollutedLocationId = Guid.NewGuid()
+        };
+
+        var response = _mapper.Map<CleaningEventCreateRequest, CleaningEvent>(createModel, businessLogicObject);
+        Assert.Multiple(() =>
```
6ar8nas: usually the structure of unit tests are:
- Arrange
- Act
- Assert

at least in my workspace, we usually try to separate all of these parts with new lines, if that's possible and/or logical. we could add a new line here as well.
```suggestion
        var response = _mapper.Map<CleaningEventCreateRequest, CleaningEvent>(createModel, businessLogicObject);
        
        Assert.Multiple(() =>
```

Apsitvarkom.UnitTests/Models/Mapping/CleaningEventMappingTests.cs
```diff
@@ -58,5 +58,33 @@ public void CleaningEventToCleaningEventResponse()
...
+        };
+
+        var createModel = new CleaningEventCreateRequest
+        {
+            StartTime = new DateTime(2022, 9, 16, 21, 43, 31).ToUniversalTime(),
+            Notes = "Hello world",
+            PollutedLocationId = Guid.NewGuid()
+        };
+
+        var response = _mapper.Map<CleaningEventCreateRequest, CleaningEvent>(createModel, businessLogicObject);
```
6ar8nas: in all our mapping tests we check the mapping behavior without colliding two objects (for example as done here - using some `default` instance).
let's change this to also use only `_mapper.Map<CleaningEvent>(cleaningEventCreateRequest);`

6ar8nas (changes requested):
tests reviewed as well.

Apsitvarkom.Api/Controllers/CleaningEventController.cs
BaVilius: I second the second option, was thinking the same while doing Update

XxDoMaSxX: All other conversations except this are resolved. Lets wait for @BenasB input about this solution.

BaVilius: You can check https://github.com/vu-vibedosa/apsitvarkom-be/pull/132 to see how I approached it, this closely resembles what @6ar8nas  advised in the second point I think

BenasB: I agree with the purpose of this comment, some thoughts:
- Probably a [400](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400) instead of a 404
- ~~Second option is the way to go IMO, just not creating `ParentExistsByPropertyAsync` on `IRepository` but on `ICleaningEventRepository` (like we have with `IPollutedLocationRepostory`), then we will not need to throw `NotSupportedException` for polluted location repository. If we can have something determined at compile time, we should go for it.~~
- Edit: Why not just use the existing `ExistsByPropertyAsync` on the polluted location repository? You can inject the `IPollutedLocationRepository` to `CleaningEventController` and just check if the polluted location exists with `ExistsByPropertyAsync`

6ar8nas: sure. thought injecting another repo might be not ideal, but we would essentially be doing the same, as is already done, so we could pretty much just reuse it instead. let's roll with that @XxDoMaSxX 

and yeah, 400 makes sense as well.. didn't put much thought into it at first.

BenasB: Well I was reading https://stackoverflow.com/questions/48449192/rest-http-response-code-for-missing-parent-resource and I might be incorrect for the status code, so we can stick to 404

6ar8nas: interesting. then let's have it a 404.

Apsitvarkom.Api/Controllers/CleaningEventController.cs
```diff
@@ -84,16 +87,30 @@ public async Task<ActionResult<CleaningEventResponse>> GetById([FromQuery] Objec
...
     [HttpPost("Create")]
     public async Task<ActionResult<CleaningEventResponse>> Create(CleaningEventCreateRequest cleaningEventCreateRequest)
     {
         var validationResult = await _cleaningEventCreateValidator.ValidateAsync(cleaningEventCreateRequest);
         if (!validationResult.IsValid) return BadRequest(validationResult.Errors.Select(e => e.ErrorMessage).ToList());
 
+        PollutedLocation? location;
+        try
+        {
+            location = await _pollutedLocationRepository.GetByPropertyAsync(x => x.Id == cleaningEventCreateRequest.PollutedLocationId);
```
6ar8nas: we don't need that object, we just need to make sure the PollutedLocation exists. we could instead use a 'lighter' (it only has to return a boolean instead of whole instance) method - `ExistsByPropertyAsync`, which only checks if such instance can be found in the database, but doesn't return any of its properties (because we don't really need them - we have the `PollutedLocationId` from the `CleaningEventCreateRequest`

6ar8nas: check out Delete action to see how it can be used - it's similar, only minor adjustments will be necessary.

Apsitvarkom.Models/Mapping/CleaningEventProfile.cs
```diff
@@ -22,6 +22,11 @@ private void MapRequests()
         CreateMap<CleaningEventCreateRequest, CleaningEvent>()
             .ForMember(x => x.Id, opt => opt.Ignore())
```
6ar8nas, line 23: here we specify all the properties that we preset - for example progress or spotted in `PollutedLocation`, `Id` here. Let's add the newly pre-set `PollutedLocationId` here as well.

Apsitvarkom.UnitTests/Api/Controllers/CleaningEventControllerTests.cs
```diff
@@ -293,6 +306,28 @@ public async Task Create_InvalidDataEntered_ValidationResultsInBadRequestRespons
...
+    public async Task Create_RepositoryThrows_Status500InternalServerErrorReturned()
+    {
+        var createRequest = new CleaningEventCreateRequest
+        {
+            PollutedLocationId = Guid.NewGuid(),
+            StartTime = DateTime.UtcNow.AddDays(1)
+
+        };
+
+        _pollutedLocationRepository.Setup(r => r.GetByPropertyAsync(It.IsAny<Expression<Func<PollutedLocation, bool>>>()))
```
6ar8nas: you only have a case where `GetByPropertyAsync` throws. however, we have two try/catch blocks there, so we would ideally cover both of them. 

you have to create a new test case, so that everything passes up until the very last step - setup `InsertAsync` to throw and assert that error.

You can have this already existent test to be named `Create_RepositoryThrowsAcquiringPollutedLocation_Status500InternalServerErrorReturned()`
and the new one:
`Create_RepositoryThrowsInsertingCleaningEvent_Status500InternalServerErrorReturned()`

Apsitvarkom.UnitTests/Api/Controllers/CleaningEventControllerTests.cs
```diff
@@ -222,16 +226,25 @@ public async Task GetById_RepositoryThrows_Status500InternalServerErrorReturned(
     [Test]
     public async Task Create_RepositoryInsertsOnce_CreatedAtActionResultReturned()
     {
+        var pollutedLocation = new PollutedLocation
+        {
```
6ar8nas: after the changes that I've suggested above (with the `ExistsByPropertyAsync) the creation of this will be no longer necessary, as we will check it only with a true or false.

6ar8nas (changes requested):
final adjustments requested. good work.

Apsitvarkom.Models/Mapping/CleaningEventProfile.cs
XxDoMaSxX: since the last commit, am not assigning PollutedLocationId values to cleaningEventDefaults object anymore. Lets the mapper do it's job :) . So it should be now as it is.

6ar8nas (approved):

PR #132, opened 2022-12-03
Implement Update for CleaningEvent

### LINKED ISSUE #125: Add Update endpoint for the CleaningEventController
## Description

Implement a new new `Update` method for the `CleaningEventController`, which should take the `CleaningEventUpdateRequest` as a `Patch` body.

## Steps to take

1. Take a look at the update implementation found in the `PollutedLocationController`. You can use it as a reference. 
2. Create the needed public models, its validators, adjust the cleaning event mapping profile.
3. Write unit tests to cover the implemented functionality in the controller, validator, mapper and other parts if modified.

## Acceptance criteria

You can update an instance of CleaningEvent using a patch request in swagger, with the endpoint /Update/.
The implementation is covered with Unit tests.

BaVilius:
Created an Update Method and wrote tests.
<hr>
Key question: Should we let change the PollutedLocation while updating a CleaningEvent? (**UPDATED** : We shouldn't)

Closes #125


BenasB:
Closes #125

BenasB:
About `ParentExistsByPropertyAsync` see https://github.com/vu-vibedosa/apsitvarkom-be/pull/131#discussion_r1038964617
(using PollutedLocation repository instead of `ParentExistsByPropertyAsync`)

> Should we let change the PollutedLocation while updating a CleaningEvent?

IMO, no. We can't change the location in `PollutedLocation` and expect the user to delete the polluted location and create another one. I think this should behave in the same way: if a user messes up and creates a cleaning event with the wrong polluted location, they should delete it and create a new one with the correct polluted location id

BaVilius:
> About `ParentExistsByPropertyAsync` see [#131 (comment)](https://github.com/vu-vibedosa/apsitvarkom-be/pull/131#discussion_r1038964617) (using PollutedLocation repository instead of `ParentExistsByPropertyAsync`)
> 
> > Should we let change the PollutedLocation while updating a CleaningEvent?
> 
> IMO, no. We can't change the location in `PollutedLocation` and expect the user to delete the polluted location and create another one. I think this should behave in the same way: if a user messes up and creates a cleaning event with the wrong polluted location, they should delete it and create a new one with the correct polluted location id

So CleaningEventUpdateRequest should only contain Id, StartTime and Notes?

BenasB:
> So CleaningEventUpdateRequest should only contain Id, StartTime and Notes?

Yes, that's how I see it: Id to identify which cleaning event to update and the other fields for actually updating the information.

Unless someone thinks we would benefit from allowing to change the polluted location as well, of course, open for discussion

P.S. I did not have the time to look into the PR changes now, but just wanted to comment on the PR description

BaVilius:
> P.S. I did not have the time to look into the PR changes now, but just wanted to comment on the PR description

That's good, I need to do some changes before asking for a review then

Apsitvarkom.Api/Controllers/CleaningEventController.cs
```diff
@@ -11,18 +11,24 @@ namespace Apsitvarkom.Api.Controllers;
 [Route("/api/[controller]")]
 public class CleaningEventController : ControllerBase
 {
-    private readonly IRepository<CleaningEvent> _repository;
+    private readonly IRepository<CleaningEvent> _cleaningEventRepository;
+    private readonly IPollutedLocationRepository _pollutedLocationRepository;
```
6ar8nas: this PR doesn't require PollutedLocationRepository to be introduced - I believe it is a left-over after the initial implementation. revert the _cleaningEventRepository name as well.

Apsitvarkom.Models/Public/CleaningEventUpdateRequest.cs
```diff
@@ -0,0 +1,21 @@
...
+
+    public string? Notes { get; set; }
+}
+
+public class CleaningEventUpdateRequestValidator : AbstractValidator<CleaningEventUpdateRequest>
+{
+    public CleaningEventUpdateRequestValidator()
+    {
+        RuleFor(l => l.Id).NotNull();
+        RuleFor(l => l.StartTime).NotNull().GreaterThan(DateTime.UtcNow);
```
6ar8nas: we should be able to have start time as null - let's take an update request, where we only want to update notes as an example.

Apsitvarkom.Models/Mapping/CleaningEventProfile.cs
```diff
@@ -19,6 +19,12 @@ private void MapRequests()
     {
         CreateMap<ObjectIdentifyRequest, CleaningEvent>(MemberList.None)
             .ForMember(x => x.Id, opt => opt.MapFrom(x => x.Id));
+        CreateMap<CleaningEventUpdateRequest, CleaningEvent>()
+            .ForMember(x => x.Id, opt => opt.PreCondition(src => src.Id is not null))
```
6ar8nas: we already check this precondition in the validator. we are sure, that the `CleaningEventUpdateRequest` has an id, so I don't think is really necessary.
However, I also don't think this is a mistake, so whatever you decide is good [it's just that I've noticed the `PollutedLocationUpdateRequest` doesn't have this, while having stuff match is preferred]

6ar8nas (changes requested):
some minor changes. didn't have the time to look at the tests yet.

Apsitvarkom.UnitTests/Api/Controllers/CleaningEventControllerTests.cs
```diff
@@ -8,6 +8,7 @@
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
 using Moq;
+using NUnit.Framework.Constraints;
```
6ar8nas:
```suggestion
```

6ar8nas: seems like an unused reference

Apsitvarkom.UnitTests/Api/Controllers/CleaningEventControllerTests.cs
```diff
@@ -214,6 +217,111 @@ public async Task GetById_RepositoryThrows_Status500InternalServerErrorReturned(
     }
     #endregion
 
+    #region Update tests
+    public async Task Update_NullIdEntered_ValidationResultsInBadRequestResponseReturned()
```
6ar8nas, line 221: `[Test]` attribute is missing - we don't ever test this method with the current setup.

Apsitvarkom.UnitTests/Api/Controllers/CleaningEventControllerTests.cs
```diff
@@ -214,6 +217,111 @@ public async Task GetById_RepositoryThrows_Status500InternalServerErrorReturned(
...
+        var resultEvent = result.Value as CleaningEventResponse;
+        Assert.That(resultEvent, Is.Not.Null);
+        Assert.Multiple(() =>
+        {
+            Assert.That(resultEvent.Id, Is.EqualTo(cleaningEvent.Id));
+            Assert.That(resultEvent.StartTime, Is.EqualTo(cleaningEvent.StartTime));
+            Assert.That(resultEvent.Notes, Is.EqualTo(cleaningEvent.Notes));
+        });
+    }
+
```
6ar8nas: Test covering a case where `UpdateAsync` throws is missing.

Apsitvarkom.UnitTests/Models/Validation/CleaningEventUpdateValidatorTests.cs
```diff
@@ -0,0 +1,62 @@
+﻿using Apsitvarkom.Models.Public;
+using FluentValidation;
+
+namespace Apsitvarkom.UnitTests.Models.Validation;
+
+public class CleaningEventUpdateValidatorTests
```
6ar8nas: looking at our code we typically have the name `'PublicModel'ValidationTests` here

6ar8nas (changes requested):

6ar8nas (approved):
looks good.

PR #134, opened 2022-12-08
Nullable properties cannot be set to null defect fix

6ar8nas:
## What was done

- Fixed an issue where we've previously had no way to set `Notes` property to null by using `Patch`.
- Additionally removed progress from PollutedLocationUpdateRequest, as this will have a separate endpoint dedicated to it.


BaVilius (approved):
lgtm

BenasB (approved):

PR #135, opened 2022-12-10
Extra validation measures, CleaningEvent adjustments

### LINKED ISSUE #126: Introduce some changes to the CleaningEvent

6ar8nas:
## What was done

- Moved mapping and validation to a separate project to avoid circular project dependency.
- Added extra validator rules for `PollutedLocationUpdateValidator`, `CleaningEventUpdateValidator`, `CleaningEventCreateValidator`
- Created a new `/Finalize/` endpoint, which should be used whenever the event is finished and we want to report the progress done.
- Introduced a `Status` property for `CleaningEvent`, which now stores the indicator whether the cleaning event's results have been already reported/finalized. 

Closes #126 + extra

Related item: #136

### DIFF
--- a/Apsitvarkom.ModelActions/Mapping/CleaningEventProfile.cs
+++ b/Apsitvarkom.ModelActions/Mapping/CleaningEventProfile.cs
@@ -31,10 +31,12 @@ public class CleaningEventProfile : Profile
     private void MapResponses()
     {
         CreateMap<CleaningEvent, CleaningEventResponse>()
-            .ForMember(x => x.Status, opt => 
-                opt.MapFrom(x => 
-                    x.IsFinalized ? CleaningEventResponse.State.Finalized : 
-                    x.StartTime > DateTime.UtcNow ? CleaningEventResponse.State.Foreseen : 
-                    CleaningEventResponse.State.Finished));
+            .ForMember(x => x.Status, opt =>
+                opt.MapFrom((cleaningEvent, _) => (cleaningEvent.IsFinalized, cleaningEvent.StartTime) switch
+                {
+                    (true, _) => CleaningEventResponse.State.Finalized,
+                    (false, var startTime) when startTime > DateTime.UtcNow => CleaningEventResponse.State.Foreseen,
+                    _ => CleaningEventResponse.State.Finished
+                }));
     }
 }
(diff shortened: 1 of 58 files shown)


Apsitvarkom.ModelActions/Mapping/CleaningEventProfile.cs
```diff
@@ -21,14 +22,18 @@ private void MapRequests()
...
     }
 
     private void MapResponses()
     {
-        CreateMap<CleaningEvent, CleaningEventResponse>();
+        CreateMap<CleaningEvent, CleaningEventResponse>()
+            .ForMember(x => x.Status, opt => 
+                opt.MapFrom(x => x.IsFinalized ? CleaningEventResponse.State.Finalized 
+                                                                        : x.StartTime > DateTime.UtcNow ? CleaningEventResponse.State.Foreseen 
+                                                                                                        : CleaningEventResponse.State.Finished));
```
BenasB: Pattern matching?

Apsitvarkom.ModelActions/Validation/CleaningEventFinalizeRequestValidator.cs
```diff
@@ -0,0 +1,52 @@
...
+                }
+
+                if (cleaningEvent.StartTime > DateTime.UtcNow)
+                    context.AddFailure("The requested cleaning event hasn't yet happened.");
+
+                if (req.NewProgress is not null && pollutedLocation.Progress > req.NewProgress)
+                    context.AddFailure($"New progress value cannot be lower than the current ({pollutedLocation.Progress}).");
+
+                return true;
+            });
```
BenasB: If you send `NewProgress = -1`, do you get both error messages? If so, should this be the expected behavior?

BenasB (approved):
A couple of questions :D haha I'm still for "clean" instead of "tidy" for error messages etc., but it's okay. Didn't look at tests too much.

Apsitvarkom.ModelActions/Validation/CleaningEventFinalizeRequestValidator.cs
6ar8nas: good catch. this is not expected, and I've just accidentally missed it. I've previously had the progress value to be constrained just to not be higher than 100 but figured out we could better have the full 0-100 shown in swagger and didn't readjust this error message. thanks.

Apsitvarkom.ModelActions/Mapping/CleaningEventProfile.cs
6ar8nas: DateTime.UtcNow is not a constant and we can have a statement body because this is a lambda expression. Will slightly adjust the indentation to make it more readable this way instead.

PR #138, opened 2022-12-10
Middleware logging requests URL, duration, status codes and total request count

### LINKED ISSUE #137: Create and use a middleware class
## Description

For the Lab#3 we need to create and use a custom middleware class. 

## Steps to take

1. Read into [what middleware is](https://www.c-sharpcorner.com/article/overview-of-middleware-in-asp-net-core/). 
2. Implement a middleware class that would track the most basic requests statistics.

## Acceptance criteria

Usage of a middleware class.

6ar8nas:
## What was done

- Created and started using a simple middleware class.

Note. If this data seems interesting, we could output it to more than just the console. This implementation is just as simple as it could have possibly been.

Closes #137

### DIFF
--- a/Apsitvarkom.Api/Middleware/LogRequestStatisticsMiddleware.cs
+++ b/Apsitvarkom.Api/Middleware/LogRequestStatisticsMiddleware.cs
@@ -5,7 +5,7 @@ namespace Apsitvarkom.Api.Middleware
     public class LogRequestStatisticsMiddleware
     {
         private readonly RequestDelegate _next;
-        private int _currentRequest = 1;
+        private int _currentRequest = 0;
 
         public LogRequestStatisticsMiddleware(RequestDelegate next)
         {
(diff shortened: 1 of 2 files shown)


6ar8nas:
image
We could also add something like this to be able to see how many times we were called during the presentation for example.

Apsitvarkom.Api/Middleware/LogRequestStatisticsMiddleware.cs
```diff
@@ -0,0 +1,34 @@
...
+        private int _currentRequest = 1;
+
+        public LogRequestStatisticsMiddleware(RequestDelegate next)
+        {
+            _next = next;
+        }
+
+        public async Task InvokeAsync(HttpContext context)
+        {
+            var requestId = _currentRequest++;
```
BenasB: I am afraid this may cause race conditions (using a primitive `int`). AFAIK, asp.net core handles requests concurrently, so bad things might happen if two threads try to access the same memory block at the same time. Using `Interlocked` should help.

BenasB (changes requested):
One comment about concurrency. We'll probably delete this after lab 3, but at least this solves the requirement for middleware 😄

BenasB (approved):

PR #146, opened 2022-12-13
Add database migrations

### LINKED ISSUE #145: Add database migrations
## Description

Up to this point, whenever any changes happen to the database schema, we always drop all the data we have previously stored in the tables. Let's start using migrations so that such development process would be lossless in terms of data.

## Steps to take

All the useful information on the topic of migrations using Visual Studio (or dotnet cli), as well as guide you can follow can be found [here](https://learn.microsoft.com/en-us/ef/core/managing-schemas/migrations/).

6ar8nas:
changed how we create the schema for the database from ensure created to migrations

## What was done

Migration process carried out following [this](https://learn.microsoft.com/en-us/ef/core/managing-schemas/migrations/?tabs=dotnet-core-cli) article.

- Added a NuGet package that is necessary for migrations.
- Created an initial migration by the name of `InitialCreate`
- Changed the way we create the db schema from calling `EnsureCreated` method before to now using `Migrate`.

Closes #145


6ar8nas:
Seems like we would need to exclude migrations from the code coverage hah

BenasB (approved):

PR #147, opened 2022-12-17
Swagger descriptions added for uncovered fluent validation rules

6ar8nas:
## What was done

- Changed the OpenApi title to `RESTful API` instead of `REST API`, because that's more accurate with what we're doing.
- Added descriptions for params in endpoints that used more complex validation rules which aren't covered by `MicroElements.Swashbuckle.FluentValidation` NuGet capabilities.

### DIFF
--- a/Apsitvarkom.Api/Apsitvarkom.Api.csproj
+++ b/Apsitvarkom.Api/Apsitvarkom.Api.csproj
@@ -5,6 +5,15 @@
     <Nullable>enable</Nullable>
     <ImplicitUsings>enable</ImplicitUsings>
     <UserSecretsId>e0e24a99-2f07-4134-a8ca-1b990f6e1a1e</UserSecretsId>
+    <GenerateDocumentationFile>True</GenerateDocumentationFile>
+  </PropertyGroup>
+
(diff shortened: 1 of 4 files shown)


Apsitvarkom.Api/Apsitvarkom.Api.csproj
```diff
@@ -5,6 +5,15 @@
     <Nullable>enable</Nullable>
     <ImplicitUsings>enable</ImplicitUsings>
     <UserSecretsId>e0e24a99-2f07-4134-a8ca-1b990f6e1a1e</UserSecretsId>
+    <GenerateDocumentationFile>True</GenerateDocumentationFile>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <NoWarn>1701;1702;1591</NoWarn>
```
6ar8nas, line 12: [regarding the first two](https://github.com/dotnet/roslyn/issues/19640) warnings (they came by default).
[CS1591](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs1591) - warning which is not important to us, as we're using the XML docs only for those methods that require additional validation information.

BenasB (approved):

PR #148, opened 2022-12-17
Add missing triggers and check constraints for database

### LINKED ISSUE #136: Add missing triggers and check constraints for database
## Prerequisites

Wait for the #135 to be completed.

## Description

Currently we only validate data that we send to the database. We should also add those constraints in the database.

## Steps to take

1. Go through the validators that we currently use and check if we have already written check constraints and/or triggers for all the values. If you find any missing - write it as is done in the `PollutedLocationContext.cs`
2. Test it locally (you can use `Adminer` for that - http://localhost:8002/)
3. Change currently added fake polluted location and cleaning event instances to comply with the rules.

## Acceptance criteria

Database no longer accepts invalid data.

XxDoMaSxX:
## What was done

- added triggers to the db


XxDoMaSxX:
can not event test if triggers working correctly. Syntax is good, but the problem is that when i update-database by initing migration or even when i paste this sql request into adminer code block i got the same error that tables are not existing. In swagger and adminer I see that those tables exist. spent like 3hrs on this error and have not found the solution (: ? Maybe anyone has any ideas ? :/

6ar8nas:
Changed the missing parts to now be correct. @XxDoMaSxX, could you please make sure all of our valid that we can do in front-end are valid here as well? Might have made some logic error possibly, although I doubt it.

Apsitvarkom.DataAccess/Migrations/20221218115321_DatabaseTriggers.cs
```diff
@@ -0,0 +1,342 @@
...
+                   "end if; " +
+                   "return new; " +
+                "end; $$ " +
+                "language plpgsql; " +
+
+                "create trigger Update_PollutedLocationAlreadyCleanedUp " +
+                "before update on \"PollutedLocations\" " +
+                "for each row " +
+                "execute procedure Update_PollutedLocationAlreadyCleanedUp();"
+            );
```
BenasB: Why not multiline string instead of concat?

https://stackoverflow.com/a/1100265/13135665

BenasB (approved):
One suggestion

